Issue #90, opened 2014-07-01
Limited Context tries to instantiate invalid plugins


dietzc:
When loading TIFF files using the standard SCIFIO `Context` (which does not include ImageJ services), SCIFIO needs the list of `Codec` objects, which triggers a call to `ObjectIndex.get(Codec.class)`, which attempts to resolve _all_ pending objects. More specifically, in our setting this includes all `net.imagej.tool.Tool` plugins, many of which rely on ImageJ-specific services not present in a SCIFIO-only `Context`.

PR #91, opened 2014-07-01
Better lazy resolution

ctrueden:
**MERGE ONLY IN CONCERT WITH THE OTHER 3.0.0 BRANCHES.**

This improves the `ObjectIndex` lazy object resolution feature to only resolve objects as needed based on the type. Previously, as soon as `get(Class)` or `getAll()` was called, _all_ pending objects would be resolved. Now we call the new `LazyObjects#getType()` method to see if the type of the `Collection` is even worth bothering about.

See also issue #90.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
 	</parent>
 
 	<artifactId>scijava-common</artifactId>
-	<version>2.41.1-SNAPSHOT</version>
+	<version>3.0.0-SNAPSHOT</version>
 
 	<name>SciJava Common</name>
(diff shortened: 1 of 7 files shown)


ctrueden:
While rebasing this branch, I noticed a problem with `AbstractSingletonService` as written. It now has a hard map of instances (due to 1b245c165d7ad40c8945070e5052005dc67d0f6b). I think we should reconsider this—perhaps the cache should be weak and lazy instead. And unit tests, unit tests.

PR #110, opened 2014-08-01
Rich plugin usage

ctrueden:

This makes all `RichPlugin` types implement the interfaces necessary for tracking their usage, and in particular tracks all uses of "handler" and "wrapper" plugin types.

Unfortunately, to do this, a method had to be removed from the `Gateway` interface, which breaks backwards compatibility. So this PR will be part of SJC3.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
 	</parent>
 
 	<artifactId>scijava-common</artifactId>
-	<version>2.28.1-SNAPSHOT</version>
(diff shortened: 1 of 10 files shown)


ctrueden:
This branch will need more scrutiny. The `UsageService` is no longer part of SJC. Probably we will want to invent a new type of event published by the `EventService`, which a new `scijava-usage` component (migrated upstream from the old `imagej-usage`) can consume.

PR #113, opened 2014-08-03
Remove CodeGenerator API

ctrueden:

It was an early exploratory attempt at code generation, and not
currently in use by the system anywhere. It will be done differently.


ctrueden:
Merged to `sjc3` integration branch with aecdd5ea6d94780a483b4e10c8a2087cf7ac3efb.

PR #140, opened 2015-01-21
Annotation ConcurrentModificationException during compilation

kmader:
Changed to ConcurrentSkipListMap to avoid compilation issue when using IntelliJ. Also added more debugging information for the AbstractIndexWriter and Reader to show which annotation causes an exception to occur so debugging can be done more easily.


ctrueden:
Merged a cleaned up version of this branch: 8f2ab8a5c406976c96dc6c7b028a7169007b7781.

PR #151, opened 2015-03-16
Improve logging

ctrueden:

Log `WARN` and `ERROR` to `stderr`, and other levels to `stdout`.

See also https://github.com/scijava/scijava-config/commit/dcf378f5fa0797179b80fe628d9ebe64dbe26621.


ctrueden:
Merged into `sjc3` integration branch.

PR #152, opened 2015-03-20
Let the ConsoleService track stdout and stderr messages

ctrueden:
This branch adds API so that interested parties can register as listeners for stdout and stderr. But not just _any_ old stdout and stderr messages—only those emitted by threads associated with the relevant SciJava application context. "Pretty snazzy," I hear you say. Well you're darn tootin' it is!


ctrueden:
@hinerm This branch is not yet ready for merge. It still needs unit tests. But I would greatly appreciate feedback on the design!

ctrueden:
Hooray, everything is working now!

Issue #109, opened 2014-07-25
Generalize ConvertService + Converter


hinerm:
The ancestors of these classes evolved to have API for handling `Type` and `Class` separately. This is unnecessary.. we should unify this API by focusing on `Type` and special-casing `Class`.

This will simplify things and reduce the number of signatures.

Furthermore the API of `<T> convert(Object, Class<T>)` is really a holdover from the old uber-conversion mindset. What we should have is:
- [x] A single high-priority (right after `NullConverter`) `castingConverter` that handles casting. This implies that low-level converters do not cast, but provide their specific output type.
- [ ] Refactor API so converters just implement `O convert(I)`

hinerm:
@dietzc raised a good question - why does [convert](https://github.com/scijava/scijava-common/blob/scijava-common-2.38.2/src/main/java/org/scijava/convert/Converter.java#L116) take an `Object` and not an `I` parameter.

Right now the answer, I believe, is to allow accepting collections and primitive arrays of type `I` as input as well. But maybe we should just always be unwrapping these multi-element inputs and relying on converting the individual elements..?

ctrueden:
> why does convert take an Object and not an I parameter.

Convenience at compile time. It is the same reason that `[Collection#contains](http://javadoc.imagej.net/Java/java/util/Collection.html#contains(java.lang.Object))` takes an `Object` instead of an `E`. I don't think there is any harm in allowing a non-`I` parameter to be specified.

ctrueden:
> maybe we should just always be unwrapping these multi-element inputs and relying on converting the individual elements

The idea was to create a separate `Converter` for collections, which recursively leans on the `ConvertService` framework for individual elements. This will reduce the SLOC count while also making conversions more robust and easier to understand.

ctrueden:
As part of this work, we should also reconsider the `canConvert(Type, Type)` style signatures. Right now, [they are deprecated](https://github.com/scijava/scijava-common/commit/01e4295c032e8fc7c646b65a148ff660c7cfb2cd), but stuff still uses them. And I think we still need them! At least, the OPS matcher does. So they should really not be deprecated...

PR #158, opened 2015-04-23
Converters

awalter17:
This is not ready to merge, because the junit tests are not running due to nested test classes.


awalter17:
This is potentially ready to merge. All of the converters' junit tests are now running and passing.

dietzc:
@awalter17 are there plans to implement converters from e.g. int[] to long[], short[] to long[], float[] to double[] etc?

ctrueden:
> are there plans to implement converters from e.g. int[] to long[], short[] to long[], float[] to double[] etc?

The only viable general solution is to be recursive. The plan is to split out the collections logic of the `DefaultConvertService` to support arrays and `Collection`s element by element. That will handle e.g. `int[]` to `long[]` as well as `int[][]` to `long[][]` etc.

This will be part of #109. See https://github.com/scijava/scijava-common/issues/109#issuecomment-78510451

dietzc:
Actually, couldn't I just write an adapter from e.g. `int[]` to `long[]` with the current framework? Or are you saying that this is not really a nice / generic way? Sorry, maybe I'm not getting your point here.

ctrueden:
@dietzc You could write such adapters, but you would need many of them. And they wouldn't handle e.g. `int[][]` to `long[][]`. Or `int[][][]` to `long[][][]`. Or... etc.

hinerm:
After testing this PR a bit it looks like it needs some more work.

Since all of the new converters are the same priority as the [DefaultConverter](https://github.com/scijava/scijava-common/blob/scijava-common-2.44.2/src/main/java/org/scijava/convert/DefaultConverter.java), they are not actually picked, and thus are not being tested when going through the `ConvertService` - the `DefaultConverter` wins. A quick fix would be to set these at slightly higher priority than the `DefaultConverter` but my preference would be to see the `DefaultConverter` gutted so it doesn't do primitive conversions any more. Any functionality it retains that might use primitive conversions (e.g. primitive array conversions..? I don't remember) would just use the `ConvertService` and thus these primitive converters.

Also, it seems like true primitive types (e.g. 'int', 'float') are not actually supported. We should not have dedicated converters for these, but ensure the [getNonprimitiveClass](https://github.com/scijava/scijava-common/blob/scijava-common-2.44.2/src/main/java/org/scijava/util/ConversionUtils.java#L143) method is being used in the `converter.supports` calls for all primitive converts, so they will properly claim these conversion requests.

A simple python script demonstrating these problems:
```
# @org.scijava.convert.ConvertService cs

from java.lang import Integer
from java.lang import Double

print cs.getHandler(Integer, Double) # Case 1. This should NOT be DefaultConverter
print cs.supports(int, double) # Case 2. This should come back IntegertoDoubleConverter
```

If the functionality is stripped out of `DefaultConverter`, case 1 does not need explicit tests. Case 2 should have unit tests, however.

ctrueden:
Indeed, removing primitive conversion from the `DefaultConverter` was supposed to be part of this work.

> case 1 does not need explicit tests.

Why not? Would be good to ensure behavior doesn't break in the future. I would prefer a unit test of it.

hinerm:
@ctrueden OK so my original tests were flawed because it was testing `PyInt` and `PyFloat` instead of java native classes. Updated to a groovy script. As a side note, the `DefaultConverter` **does** properly support primitive conversion (as we inferred from looking at the unit tests)

>>    case 1 does not need explicit tests.

>Why not? 

I meant that we may not want the unit tests to rely on particular concrete classes coming back as handlers, but we *do* want to verify the convert service can handle primitive conversions. Or at least these tests shouldn't be conflated (i.e. I think a test for for `ConvertService` is the most important.. but maybe a test for primitive converters that has overlap but uses concrete converters would be useful..?)

hinerm:
OK I just added 2 commits that allow the new primitive converters to handle unboxed primitive conversion requests. Currently they still are not because of their priority relative to the `DefaultConverter`.. but as the `DefaultConverter` is gutted primitive conversion should remain functional.

ctrueden:
Just want to point out a use case where having the `CastingConverter` might be really nice: imagej/imagej-ops@da4b24e09cea7f7a470d35402ea466252a702b8c. Because as things stand, that code actually ignores generic types for the most part, so e.g. an `ArrayList` argument (which was written as `ArrayList<String>` and contains only `String` objects) would be matched with a `Collection<? extends Class<?>>` since the `Type` describing the latter boils down to a raw `Collection` and nothing more...

ctrueden:
Rebased over latest master, and all tests pass. So let's merge this long overdue branch.

PR #163, opened 2015-04-30
Script values

ctrueden:
This branch makes default values work better:
* It gives the `ModuleItem` a default value property which can be retrieved and (in the case of `MutableModuleItem`) set.
* It uses that property for the default value, via a new `DefaultValuePreprocessor`. This eliminates the inconsistency between default value handling of primitives and their `java.lang` wrapper types.
* It adds a `value` attribute to script parameter parsing, so that you can write e.g. `@int(value=5) sigma` to set a default value for script parameters. (Ideally, we'd support `@int sigma = 5` too, but for the moment that would be more work, so we can punt for now.)


ctrueden:
@hinerm Just looking for quick code review, then feel free to merge!

hinerm:
Awesome @ctrueden ! :octopus:

PR #164, opened 2015-05-04
Add API to give explicit file chooser title

hinerm:
The method `UserInterface#chooseFile(String, File, String)` is given a deprecated default implementation in `AbstractUserInterface` to avoid requiring a major version bump according to SemVer.


ctrueden:
I tweaked this commit to avoid requiring a major version bump. So we can merge now!

PR #165, opened 2015-05-08
Use generic types in the ConvertRequest class

ctrueden:
A converter may need to verify type compatibility not just at the
Class level, but potentially for any generic type. Generic types are
not available at runtime from object instances, but they are
available in other scenarios; e.g., the ImageJ OPS framework uses the
converter framework to decide type compatibility (i.e., assignability
or convertibility) based on generic types. This is possible (rather
than generics being erased at runtime) because OPS parameters are
declared as instance fields, which can be introspected at runtime.


hinerm:
:octopus:

hinerm:
(translation: I require 8 arms to express my happiness with this PR)

PR #169, opened 2015-06-04
Set locale to US so that the strings are always formated as expected.

gab1one:
In European locales "," is used as decimal seperator instead of ".". String.format(..) automatically converts decimal numbers,
making some unit tests fail on systems with certain European locales.


hinerm:
Awesome, thank you @gab1one!

PR #170, opened 2015-06-08
Locale fixes

hinerm:
Makes UnitUtils non-US-centric, and updates the corresponding unit test to pass in non-US locales.

### DIFF
--- a/src/main/java/org/scijava/util/UnitUtils.java
+++ b/src/main/java/org/scijava/util/UnitUtils.java
@@ -66,9 +66,14 @@ public final class UnitUtils {
 		// compute value from unit
 		final double value = totBytes / Math.pow(1024.0, pow);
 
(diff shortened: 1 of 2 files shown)


hinerm:
@gab1one Would you mind checking out this branch and building it on your machine? It changes the fix you submitted last week to not force US Locale, but instead format the test strings for the current locale.. just want to be sure the unit tests don't fail for you now. Thanks!

gab1one:
Sadly that does not work at all (the local() method has no effect), I think we need to split the number from the unit or something.

hinerm:
>I think we need to split the number from the unit or something.

@gab1one yep I agree. When the test strings contain the numeric value + label, they don't get numeric formatting.

Anyway I force pushed a new patch set that should format the test strings the same as UnitUtils. Can you test again?

Much thanks.

gab1one:
It works now, thank you :+1:

PR #173, opened 2015-06-25
ThreadService: add API for backing ExecutorService

ctrueden:
For some uses cases (particularly for certain calls to ImgLib2 functionality), it is helpful to have access to the ExecutorService being used internally by the ThreadService. And while we're at it, we may as well make the ExecutorService itself configurable.

I have a vague intuition that ExecutorService is probably not applicable to all threading models of all possible sorts of Java runtimes, so I include a caveat in the javadoc that the accessor is allowed to return null, and the mutator is allowed to throw UnsupportedOperationException, if the ThreadService implementation does not use an ExecutorService.

### DIFF
--- a/src/main/java/org/scijava/thread/DefaultThreadService.java
+++ b/src/main/java/org/scijava/thread/DefaultThreadService.java
@@ -84,6 +84,16 @@ public final class DefaultThreadService extends AbstractService implements
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


ctrueden:
@dietzc Happy Thursday.

dietzc:
thanks so much @ctrueden

PR #178, opened 2015-08-21
Add array conversion classes

hinerm:
Adds converters for going between analogous primitive array (and list) types.

### DIFF
--- a/src/main/java/org/scijava/convert/ArrayConverters.java
+++ b/src/main/java/org/scijava/convert/ArrayConverters.java
@@ -1,0 +1,446 @@
+/*
+ * #%L
+ * SciJava Common shared library for SciJava software.
+ * %%
+ * Copyright (C) 2009 - 2015 Board of Regents of the University of
+ * Wisconsin-Madison, Broad Institute of MIT and Harvard, and Max Planck
+ * Institute of Molecular Cell Biology and Genetics.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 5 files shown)


hinerm:
Needs unit tests.

hinerm:
@ctrueden @bnorthan - updated and working now, I hope.

bnorthan:
@hinerm 

It is still not working in python, I think it because 'python lists' don't get converted to 'Dimensions'.  'Python lists' do get converted to 'java arrays', and java arrays get converted to 'Dimensions', but 'python lists' don't get converted to 'Dimensions'

``` python
# this one works...
phantom=ops.create().img(array([xSize, ySize], 'l'))

# so does this one...
phantom=ops.create().img(FinalDimensions([xSize,ySize]))

# this one does not....
phantom=ops.create().img([xSize, ySize])
```

hinerm:
Yeah there were several fixes required at various layers.

>phantom=ops.create().img([xSize, ySize])

This is addressed https://github.com/imagej/imagej-ops/commit/becc4904fb639f6835a127a7d9e30112f1c593fb should fix

hinerm:
>>'Python lists' do get converted to 'java arrays'

>This requires a separate converter, which is incoming! Then this will work:

I lied! I thought you were referring to failures when passing an array + another argument, e.g. a `Type`, which wasn't working. But actually this already works without the need for a custom converter, because of the `List > FinalDimensions` converter:

```
ops.create().img([xSize, ySize], LongType())
```

PR #194, opened 2015-10-08
Independent Readers in ScriptInfo

Squareys:
Hi @ctrueden !

here is the pullrequest promised in https://github.com/scijava/scripting-jython/issues/6. My solution changes `ScriptInfo` as follows:

*Instead of `ScriptInfo.getReader()` returning the reader passed onto the constructor of ScriptInfo, the constructor now reads the entire contents of the Reader into a string, so that `ScriptInfo.getReader()` can return a new `StringReader`.*

Greetings, 
Squareys

### DIFF
--- a/src/main/java/org/scijava/script/ScriptInfo.java
+++ b/src/main/java/org/scijava/script/ScriptInfo.java
@@ -36,6 +36,7 @@ import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
 import java.io.Reader;
+import java.io.StringReader;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.HashMap;
@@ -73,7 +74,7 @@ public class ScriptInfo extends AbstractModuleInfo implements Contextual {
 	private static final int PARAM_CHAR_MAX = 640 * 1024; // should be enough ;-)
 
 	private final String path;
-	private final BufferedReader reader;
+	private final String script;
 
 	@Parameter
 	private Context context;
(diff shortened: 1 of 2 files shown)


src/main/java/org/scijava/script/ScriptInfo.java
```diff
@@ -478,4 +489,27 @@ else if ("value".equalsIgnoreCase(key)) {
...
+	 * @param reader {@link Reader} whose contents should be returned as String.
+	 *          Expected to never be <code>null</code>.
+	 * @return contents of reader as String.
+	 * @throws IOException If an I/O error occurs
+	 * @throws NullPointerException If reader is <code>null</code>
+	 */
+	private static String getReaderContentsAsString(final Reader reader)
+		throws IOException, NullPointerException
+	{
+		final char[] buffer = new char[8192];
```
Squareys: This number is pretty arbitrary. Any ideas on that? Use the probably also arbitrary (but alot bigger) `PARAM_CHAR_MAX` instead? :)

This code is based off of the code I wrote in scripting-java a while ago.

Squareys:
It would be great, if someone could quickly review this, since this would fix https://github.com/scijava/scripting-jython/issues/6.

ctrueden:
Thanks @Squareys. In retrospect, the design of the `ScriptInfo` is pretty bad. I'm sorry about that. Your changes are a nice patch to the issue. It is better to just work with `String` always, and forget the whole `Reader` thing, since script size does not need to scale beyond very small values.

The `8192` value is fine for now. Perhaps some day SJC will depend on Guava and then we can call a utility method (`CharStreams.toString`) to read the data in a more succinct way.

PR #203, opened 2015-11-13
Make FileUtils.stripFilenameVersion support jogamp

ctrueden:
Actually, it is more general than that: it identifies native platform suffixes more generally now, which will hopefully work with more than just the jogamp platform artifacts.

But we do add a test explicitly for jogamp's jogl-all artifacts, since those are the ones we acutely need to support.

@hinerm Could you please do a quick review?


hinerm:
Awesome changes! :+1:

ctrueden:
Thanks for looking at it!

PR #206, opened 2015-11-20
DefaultModuleService: do not persist default value

ctrueden:
@LeonYang5114 Please test, and let me know what you think.

### DIFF
--- a/src/main/java/org/scijava/module/DefaultModuleService.java
+++ b/src/main/java/org/scijava/module/DefaultModuleService.java
@@ -60,6 +60,7 @@ import org.scijava.service.AbstractService;
 import org.scijava.service.Service;
 import org.scijava.thread.ThreadService;
 import org.scijava.util.ClassUtils;
+import org.scijava.util.MiscUtils;
 
 /**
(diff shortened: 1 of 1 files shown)


LeonYang5114:
@ctrueden when I was debugging, the `save` method is never called as I run the scripts to get a numeric input, and I think the case that `value == item.getDefaultValue()` is unlikely, since the [LoadInputsPreprocessor](https://github.com/scijava/scijava-common/blob/master/src/main/java/org/scijava/module/process/LoadInputsPreprocessor.java#L84) will set the the value to a non-null default value.

Now that I understand how the PerfService works, I think what causing the problem might be that:
1. We might not always want to persist parameters from the previous session. It would be helpful to have a checkbox for the user to choose the behavior, and a reset button, as you mentioned.
2. If we choose to persist the value from the last session, I think we should first check whether the range and type are consistent, even in the same script.  If we change the valid range of a parameter in a new session, the persistent value in the old session might be out of range, so that the spinner will not work.

Hope that this is clear.

ctrueden:
Merging, since I don't think it really hurts anything. We can revert later if I am wrong.

Issue #207, opened 2015-11-20
Default value for a numeric @Parameter is ignored


LeonYang5114:
A `@Plugin` with the following `@Parameter` is supposed to have 10 as the input's default value (if no persist value exists), but `min` is always used instead.
```
	@Parameter(min = "1", max = "50")
	private double test = 10;
```

The problem roots in the [DefaultValuePreprocessor#assignDefaultValue](https://github.com/scijava/scijava-common/blob/35163f8e93e526ab75f710f014f4600ded4398b4/src/main/java/org/scijava/module/process/DefaultValuePreprocessor.java#L73), which sets any numeric values to min/max/zero as long as the variable is not "resolved".

Since the statement `private double test = 10;` effectively resolved the variable, the DefaultValuePreprocessor should not overwrite the default value.

I think a procedure for the DefaultValuePreprocessor should be:
1. Check resolved. If resolved, return.
2. Check validity. If already valid, return. For numeric type, we should check within range.
3. Set default value as `getDefaultValue()` if the returned value is not null.

PR #210, opened 2015-11-25
Fix bug of overwriting pre-assigned values

LeonYang5114:
The DefaultValuePreprocessor overwrote the value of the module's
corresponding item even if the item is pre-assigned with some value. Now
it checks if the item is primitive or if it is non-null before setting
the default value.

See also:
#207


LeonYang5114:
My commit is a quick and dirty solution.

The ideal way to solve this problem would be having `moduleService.getDefaultValue(item)` return the pre-assigned values, but I don't think right now we can achieve this since each `ModuleItem` only has information about the annotation. We could have a `value` or similar attribute, or hack the `PluginInfo` so that it can read the initial values of the parameter, or hope anyone can come up with a better solution.

ctrueden:
Thanks! I actually think this solution is fine in general, and not too hacky.

PR #211, opened 2015-11-25
Add empty option for not-required parameter in Widget

LeonYang5114:
For command parameter that is not required but candidate options exist,
DefaultWidgetModel gave no empty option to be chosen from, which force
the user to use the potentially invalid values.

### DIFF
--- a/src/main/java/org/scijava/widget/DefaultWidgetModel.java
+++ b/src/main/java/org/scijava/widget/DefaultWidgetModel.java
@@ -83,6 +83,8 @@ public class DefaultWidgetModel extends AbstractContextual implements WidgetMode
 		this.module = module;
 		this.item = item;
 		this.objectPool = objectPool;
+		if (!item.isRequired())
+			this.objectPool.add(0, null);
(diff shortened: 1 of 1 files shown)


LeonYang5114:
@ctrueden As for the null value problem, it seems that you already assume its validity, since you already [check null](https://github.com/scijava/scijava-common/blob/45ab6cf22dc3c33400bc40a17885e968f4eaa8ad/src/main/java/org/scijava/widget/DefaultWidgetModel.java#L235-L237) in the `getText()` method. As it is, I think it might not be too bad to add a null to the object pool and set its text as "None" or "-------None-------" etc, as you mentioned.

ctrueden:
The null check in the widget model code is only because the value can _start_ as null. But once the user has made any changes to the widget, it will no longer be so, _ever_.

ctrueden:
I think this is probably a step in the right direction. I think it is very likely that it does not fully solve the use of nulls for optional parameter widgets, though. But I guess we can merge and go from there.

PR #212, opened 2015-12-02
Add @Override annotation to method missing it.

gab1one:
Just a tiny fix.


ctrueden:
I was avoiding changes to the `MerseinneTwisterFast` because it is a copy/paste from elsewhere, and changing it makes it harder to upgrade later. That's why the code style does not match the rest of SciJava, either. But I guess it's not really a big deal in this case.

PR #214, opened 2015-12-06
Only try to load services if required.

gab1one:
If the serviceClasses Collection is empty, there is no reason to create a ServiceHelper.

This is a source of nullpointers in the ``SubContext`` and a tiny performance improvement when creating lots of ``SubContext`` instances.

### DIFF
--- a/src/main/java/org/scijava/Context.java
+++ b/src/main/java/org/scijava/Context.java
@@ -272,9 +272,11 @@ public class Context implements Disposable {
 
 		setStrict(strict);
 
-		final ServiceHelper serviceHelper =
-			new ServiceHelper(this, serviceClasses, strict);
-		serviceHelper.loadServices();
+		if (!serviceClasses.isEmpty()){
+			final ServiceHelper serviceHelper =
+					new ServiceHelper(this, serviceClasses, strict);
+			serviceHelper.loadServices();
+		}
 	}
 
 	// -- Context methods --


ctrueden:
Shouldn't we fix the NPE inside the `ServiceLoader` too though?

gab1one:
The problem was at [this](https://github.com/gab1one/scijava-common/blob/master/src/main/java/org/scijava/service/ServiceHelper.java#L270) line. My implementation returned null as the ServiceIndex. I fixed that by returning an empty ServiceIndex, I just think if we stay true to the __fully empty context__ idea, not trying to load makes a lot of sense.

ctrueden:
> I just think if we stay true to the fully empty context idea, not trying to load makes a lot of sense.

It shouldn't matter, and it adds complexity to the code. I am against it for the same reason I would be against code like:
```
final List<T> myList = getList();
if (!myList.isEmpty()) {
    for (final T myItem : myList) {
        doStuff(myItem);
    }
}
```
When you could just as easily use the simpler:
```
for (final T myItem : getList()) {
    doStuff(myItem);
}
```
(as long as you know the list won't be `null` of course)

ctrueden:
> a tiny performance improvement

And I don't believe in "tiny performance improvements" unless you actually measure them.

gab1one:
Ok currently I get this warning when I use the [new constructor](https://github.com/scijava/scijava-common/blob/master/src/main/java/org/scijava/Context.java#L130): 
``Class pool is empty: forgot to call Thread#setClassLoader``.

Which makes sense, as it tries to load an empty list. So I am not sure how we can avoid this without adding a check somewhere, either in the ``ServiceHelper`` or in the ``Context`` itself.

ctrueden:
Sorry @gab1one, your non-perma-link into master got skewed as time went on. So I am not actually sure what your question is. I am inclined to close this PR. Unless there is still a problem you need to resolve here?

gab1one:
The problem is still there:

If you use [this  constructor](https://github.com/scijava/scijava-common/blob/fd8a932346695cd0a156a994bae74b9c0c22ed39/src/main/java/org/scijava/Context.java#L130) with the arguments ``false, false``.The servicehelper will log the warning: ``Class pool is empty: forgot to call Thread#setClassLoader?.`` [at this point](https://github.com/scijava/scijava-common/blob/696913021a021d21d6617e82a90799ad31e76a06/src/main/java/org/scijava/service/ServiceHelper.java#L122).
The reason is that the [findServiceClasses()](https://github.com/scijava/scijava-common/blob/696913021a021d21d6617e82a90799ad31e76a06/src/main/java/org/scijava/service/ServiceHelper.java#L359) method fails to find any service classes, because as the context does not contain any services. 
It makes little sense to have a constructor that logs warning messages when used as intended. I still believe that skipping the service loading when there are no services to load is the correct way to handle this issue.

PR #216, opened 2016-01-17
Allow context multi-injection in non-strict mode

ctrueden:
With `scijava.context.strict` set to `false` (i.e.: `strict=false` flag of the `Context`), injecting an object multiple times is now allowed.

This avoids the dreaded "Context already injected" exception which is thrown, for example, when reusing an op with a second module.


ctrueden:
@dietzc If you get a chance, feel free to test whether this makes your life any easier.

You just have to pass `strict=false` when spinning up your `Context`s. Or set the `scijava.context.strict` system property to `false` for all of them to be non-strict by default.

PR #217, opened 2016-01-18
Use List instead of Iterable as returntype of ModuleInfo methods.

gab1one:
that return a list of ModuleItems.

``inputs()`` and ``outputs()`` used to return ``Iterable<ModuleItem<?>> ``this made random access on a Module's inputs and outputs painful. This restriction is not needed as the implementations of ModuleInfo already return a List. This change breaks compatibility with custom implementations of ``ModuleInfo``, but not with any calls to the changed methods, as ``List`` is still ``Iterable``.

### DIFF
--- a/src/main/java/org/scijava/command/CommandInfo.java
+++ b/src/main/java/org/scijava/command/CommandInfo.java
@@ -279,13 +279,13 @@ public class CommandInfo extends PluginInfo<Command> implements ModuleInfo {
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


ctrueden:
Thanks, this is long ([looooooong](https://www.youtube.com/watch?v=rvramBC_P_U)) overdue!

ctrueden:
Reverted in edfc7c53815918672ba708daa5fb45b3d80ea404 :sob:

PR #218, opened 2016-01-19
Update annotation preprocessor to support java8

gab1one:
This change allows library users to write SciJava Plugins using java8 language features without getting constant warnings and errors from annotation preprocessor.

The commit with the tests: https://github.com/scijava/scijava-common/commit/a9ba97e40ddc642c182c7f0a0098136795300ea7 can't get merged without scijava-common using java8. Everything else works also with java6.

### DIFF
--- a/src/main/java/org/scijava/annotations/ByteCodeAnalyzer.java
+++ b/src/main/java/org/scijava/annotations/ByteCodeAnalyzer.java
@@ -107,6 +107,8 @@ class ByteCodeAnalyzer {
 			getU4(offset + 5));
 	}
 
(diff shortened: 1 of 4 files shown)


ctrueden:
Thank you so much! That limitation had really been annoying me too.

ctrueden:
So: to merge this, you just need to add `testTarget` and `testSource` set to `1.8`. Then verify it all works fine on the CLI, and that Java 6 can consume the main JAR (but not the test JAR, of course) afterward.

We cannot upgrade SJC as a whole to Java 8 until ImageJ has thoroughly switched at the higher levels first.

PR #220, opened 2016-01-25
Add a shiny new REPL for the script framework

ctrueden:
Well, a little shiny. It's pretty basic, but the language switching feature is pretty snazzy. And it's convenient to have for CLI usage.


ctrueden:
I updated ImageJ's Swing-based script interpreter plugin to use this new `ScriptREPL` instead, and everything went off without a hitch. It will be possible to expand on the functionality fairly easily as well. I am happy with this work now.

PR #222, opened 2016-02-03
Add support for custom main classes

ctrueden:
It is nice to be able to override which main class gets executed.
Actually, the plan is not to "override" it per se, but actually for
whichever main application class to have the option to call
mainService.execMains() to run the main classes which have been
registered (via the "--main" CLI argument or otherwise via the API).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
 	</parent>
 
 	<artifactId>scijava-common</artifactId>
-	<version>2.50.2-SNAPSHOT</version>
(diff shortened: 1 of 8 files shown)


hinerm:
@ctrueden I like it, a lot actually. My only thought: add a test that calls two different trivial main methods?

ctrueden:
Thanks for the reminder. Adding tests now.

ctrueden:
OK, tests added. Issues were discovered and fixed as a consequence.

hinerm:
:+1:

PR #224, opened 2016-02-05
Reintroduce ModuleInfo List iterators

ctrueden:
__INTENDED FOR MERGE WITH OTHER SJC3 WORK__

In 88d3cab189d1fd68b2065e0e109cbb18ef48e0b0, the type of the iterators returned by inputs() and outputs() was narrowed from Iterable to List. Unfortunately, this broke downstream callers, so the change was reverted in edfc7c53815918672ba708daa5fb45b3d80ea404.

This commit reintroduces the patch, for inclusion into SJC3.


ctrueden:
Merged with `sjc3` integration branch with 2543f695f7528643d506794945eca4dd270d006d.

PR #229, opened 2016-02-23
Fix horrible bug in Command cancelation

ctrueden:
If your `Command` implemented `Cancelable`, then calling `cancel(...)` from within the command itself (e.g., from an initializer callback) did not actually cancel anything, because the wrapping `CommandModule` instance did not recognize the cancelation state of its wrapped `Command`.

Now, the `CommandModule` always leans on its wrapped `Command` for all cancelation-related operations, unless the `Command` itself does not implement `Cancelable`.

Noticed by Richard Domander. Thanks to Mark Hiner for help debugging.


ctrueden:
@rimadoma Tested and confirmed that the fix works for his scenario.

PR #230, opened 2016-02-24
Do proper exception handling when loading classes

ctrueden:
Several places in SciJava Common call the `ClassUtils.loadClass` routine. Previously, that routine simply returned null without explanation when the class could not be loaded for any reason. Now, by setting the `quietly` flag to false, we can do proper exception handling so that the reason for failure is known and can be passed along to the caller.

So this branch does that in all the places where it makes sense.


ctrueden:
@EikeHeinz Sorry it took so long to address this problem. Hopefully this makes the arcane class loading failures a little more transparent in the future. Release coming soon...

PR #237, opened 2016-03-29
Improve parsing of script attributes

ctrueden:
Among many other benefits, this enables support for the choices attribute, so that scripts can have multiple choice parameters.


ctrueden:
@hinerm Could you please review, when you have a moment? Everything is tested and working, but of course it's always good to be careful when mucking around in SJC.

hinerm:
:+1: looks sane to me @ctrueden

PR #238, opened 2016-04-13
Add support for arbitrary parameter attributes

ctrueden:
It was not possible to specify `@Attr` key/value pairs in scripts.
With this change, it is now not only possible, but nice and elegant!


ctrueden:
@gab1one Please test! 😄

ctrueden:
From [Gitter](https://gitter.im/scijava/scijava-common/archives/2016/04/14):
> @gab1one 05:57
> It looks great, thank you!

PR #242, opened 2016-06-19
Improve behavior of return value outputs

ctrueden:
When _no_ explicit `@OUTPUT` is given, the return value _is_ appended.

When an explicit `@OUTPUT` _is_ given, the return value is _not_ appended.

See knime-ip/knip-scripting/issues/78.


ctrueden:
@gab1one Please test.

ctrueden:
@gab1one Any progress? I just rebased this over the latest master, so that it continues to merge cleanly.

ctrueden:
@gab1one Since I have still not heard back, I am going to merge this. I hope it does not break anything.

PR #243, opened 2016-06-20
Add array type aliases to the DefaultScriptService

gab1one:
This allows script users to easily use arrays with the added java
types, without needing to fully qualify the array type:

``@Integer[] input`` instead of ``@java.lang.Integer[] input``


ctrueden:
It would be nice, rather than proliferating all these array aliases, to instead add first-class support for propagation of existing aliases to the array cases. No?

ctrueden:
I did it. So this PR is now unnecessary.

PR #244, opened 2016-07-12
Adds method to get InputWidget from InputPanel

dietzc:
sometimes one wants to access the `InputWidget` from `InputPanel`s directly. Therefore one can request now the `InputWidget`from an `InputPanel`. However, this breaks backwards compatibility.


ctrueden:
Thanks! 😄

ctrueden:
Regarding backwards compatibility: no problem, because the abstract class offers the implementation. Our policy in SJC is that new interface methods can be added, as long as the corresponding abstract class (or the interface itself, thanks to Java 8 default methods) provides a default implementation. See also http://imagej.net/Versioning#Scope_of_SemVer, where we actually do not promise even that—we guarantee only _API_ compatibility, not _SPI_.

PR #245, opened 2016-08-12
Fix headless UI bug

rimadoma:
Fixed the bug where dialog would pop open even if UI was set _headless_. Sorry it took me so long to submit this bug fix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,14 @@
 *.swp
+
+# Maven #
+/target/
+
+# Eclipse #
 /.classpath
 /.project
 /.settings/
-/target/
+
+
+# IntelliJ #
+/.idea/
+/*.iml
--- a/src/main/java/org/scijava/ui/DefaultUIService.java
+++ b/src/main/java/org/scijava/ui/DefaultUIService.java
@@ -214,7 +214,6 @@ public final class DefaultUIService extends AbstractService implements
 	public UserInterface getDefaultUI() {
 		if (defaultUI != null) return defaultUI;
 		if (uiList().isEmpty()) return null;
-		if (defaultUI != null) return defaultUI;
 		return uiList().get(0);
 	}
 
--- a/src/main/java/org/scijava/ui/DefaultUIService.java
+++ b/src/main/java/org/scijava/ui/DefaultUIService.java
@@ -68,6 +68,7 @@ import org.scijava.ui.DialogPrompt.MessageType;
 import org.scijava.ui.DialogPrompt.OptionType;
 import org.scijava.ui.DialogPrompt.Result;
 import org.scijava.ui.event.UIShownEvent;
+import org.scijava.ui.headlessUI.HeadlessUI;
 import org.scijava.ui.viewer.DisplayViewer;
 
 /**
@@ -212,6 +213,7 @@ public final class DefaultUIService extends AbstractService implements
 
 	@Override
 	public UserInterface getDefaultUI() {
+		if (isHeadless()) return HeadlessUI.getInstance();
 		if (defaultUI != null) return defaultUI;
 		if (uiList().isEmpty()) return null;
 		return uiList().get(0);
--- a/src/main/java/org/scijava/ui/UserInterface.java
+++ b/src/main/java/org/scijava/ui/UserInterface.java
@@ -50,7 +50,7 @@ import org.scijava.widget.FileWidget;
  * implementing this interface, it is encouraged to instead extend
  * {@link AbstractUserInterface}, for convenience.
  * </p>
- * 
+ *
  * @author Curtis Rueden
  * @see Plugin
  * @see UIService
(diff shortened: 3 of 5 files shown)


pom.xml
```diff
@@ -126,6 +131,18 @@
...
 		</dependency>
+		<dependency>
+			<groupId>org.mockito</groupId>
+			<artifactId>mockito-core</artifactId>
+			<version>1.10.19</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.scijava</groupId>
+			<artifactId>scijava-ui-swing</artifactId>
```
ctrueden: This is a circular dependency: `scijava-ui-swing` depends on `scijava-common`. Could you please, rather than mocking a `SwingSDIUI`, just implement your own `UserInterface` plugin in the test class itself as a static inner class, and then use that? Then you wouldn't need Mockito either.

src/main/java/org/scijava/ui/DefaultUIService.java
```diff
@@ -212,9 +213,9 @@ public boolean isHeadless() {
 
 	@Override
 	public UserInterface getDefaultUI() {
+		if (isHeadless()) return null;
```
ctrueden: I really like this. It makes total sense to return `null` in a headless context. And if downstream code can't deal with it—tough. 😝

src/main/java/org/scijava/ui/DefaultUIService.java
```diff
@@ -257,16 +258,31 @@ public UserInterface getUI(final String name) {
 
 	@Override
 	public void show(final Object o) {
-		getDefaultUI().show(o);
+		final UserInterface defaultUI = getDefaultUI();
+		if (defaultUI == null) {
```
ctrueden: Please use the more compact one-line syntax:
```java
if (defaultUI == null) return;
```
for consistency with the other blocks of code in these here parts.

src/test/java/org/scijava/ui/DefaultUIServiceTest.java
```diff
@@ -0,0 +1,32 @@
...
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.mockito.Mockito.mock;
+
+import org.junit.Test;
+import org.scijava.ui.swing.sdi.SwingSDIUI;
+
+/**
+ * Tests for {@link DefaultUIService}
```
ctrueden: By convention, I have named the service tests after the interfaces. While technically maybe less correct, it produces shorter class names, and it's not like we unit test any _other_ service interface implementations.

So for consistency, let's call this `UIServiceTest` instead of `DefaultUIServiceTest`.

.gitignore
```diff
@@ -1,5 +1,14 @@
...
+/target/
+
+# Eclipse #
 /.classpath
 /.project
 /.settings/
-/target/
+
+
+# IntelliJ #
```
ctrueden: Please make the gitignore changes in a separate commit.

ctrueden:
Thank you very much, Richard. Could you please take care of my requested modifications? Then I will definitely merge it—this is a critical fix.

ctrueden:
So, my main question is: should we, rather than returning `null` in `getDefaultUI()`, instead implement a low-priority `HeadlessUI` in SciJava Common itself?

rimadoma:
A rather like the idea of a null object. A third option would be to return `Optional<UserInterface>` from `getDefaultUI()`, but that'd mean backwards incompatible changes to the API, which we don't want at this point. Shall I get on with the `headlessUI`? Should it be instantiated and returned from `getDefaultUI` if the object is running headless?

ctrueden:
> Shall I get on with the `headlessUI`? Should it be instantiated and returned from `getDefaultUI` if the object is running headless?

Sounds great. Thank you for working on it!

rimadoma:
Sorry it took so long, I was travelling. There's a lot of boiler plate code in `HeadlessUI` and its helper classes to make them null objects. Would it be OK to just return `null` values from the methods of `HeadlessUI` like in `AbstractUserInterface`?

ctrueden:
@rimadoma Yes, I think you can simply return `null` for the main UI elements. If you run into places where downstream code assumes non-null, we can fix it. The original idea was that each UI may or may not have each of the main components—e.g., no status bar or tool bar. I don't think we need to make dummy objects for these.

rimadoma:
Returns `null` objects now. Updated Javadoc of `UserInterface`. Commented overridden `context()` and `getContext()` in `HeadlessUI` since they work differently in the null object than specified in the `Contextual` interface.

ctrueden:
I am in the process of cleaning this up and enhancing it a little. Will merge soon.

ctrueden:
Whew, that took awhile. But totally worth it! Thanks again @rimadoma.

PR #247, opened 2016-09-13
Add STL format

rimadoma:
Add support for standard binary STL files, and possibility to extend i/o for ascii STLs etc.

This PR is just to initiate support for STL files, I am aware it'll probably need a lot of work to be merged. For instance, I could add support for ASCII STL files from the ThreeDViewer repo of @kephale.

What would be a good way to determine `boolean supports` in `STLIOPlugin`, i.e. when trying to decide between using binary and ascii formats for opening and saving files?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -119,6 +119,10 @@
 			<artifactId>eventbus</artifactId>
 			<version>1.4</version>
 		</dependency>
+		<dependency>
+			<groupId>com.google.guava</groupId>
+			<artifactId>guava</artifactId>
+		</dependency>
 
 		<!-- Test scope dependencies -->
 		<dependency>
--- a/src/main/java/org/scijava/stl/AbstractBinarySTLFormat.java
+++ b/src/main/java/org/scijava/stl/AbstractBinarySTLFormat.java
@@ -1,0 +1,75 @@
+
+package org.scijava.stl;
+
+import com.google.common.base.Strings;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import org.scijava.plugin.AbstractHandlerPlugin;
+import org.scijava.util.FileUtils;
+
+/**
+ * An abstract superclass of binary {@link STLFormat} implementations
+ * <p>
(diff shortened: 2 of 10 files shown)


rimadoma:
Using a custom `STLFacet` class, because I don't think I can use `Mesh`, `DefaultMesh` or some such without creating a circular Maven dependency

ctrueden:
This is great! But it should live outside SciJava Common -- perhaps in a
new scijava-plugins-io or scijava-plugins-io-mesh component. I am traveling
right now but will follow up soon with more detailed thoughts.

ctrueden:
@kephale How does this jive with your [ImportSTL](https://github.com/kephale/ThreeDViewer/blob/734330c285976ce40f0c1eefb098a6ea1295392e/src/main/java/sc/fiji/threedviewer/io/ImportSTL.java) logic etc.?

How do you guys feel about a new `imagej/imagej-plugins-io-mesh` component with support for mesh formats? Just like @rimadoma has done here, these should be SciJava `IOPlugin`s (not SCIFIO `Format` plugins) because they are not image data.

The dependencies are not an issue—we can make `imagej-plugins-io-mesh` depend on `imagej-ops` and `guava` and whatever else is needed. Then ship `imagej-plugins-io-mesh` with ImageJ, use it as a dependency in `ThreeDViewer`, etc.

Am I missing something, or will this plan be awesome?

hadim:
I was looking for a way to export dataset to stl. This PR seems awesome.

@rimadoma are you planning to move it to a new `imagej/imagej-plugins-io-mesh`  repo ?

ctrueden:
I had a few spare minutes, so I created the repository and pushed this work:

* https://github.com/imagej/imagej-plugins-io-mesh

@rimadoma You are currently listed as a contributor rather than a developer. If you want to help maintain or support this component, that would be much appreciated, and I can move your entry in the POM into the `<developers>` section. See http://imagej.net/Team for details.

@kephale I tentatively made you project lead even though your code is not in here yet, since we had previously discussed creating a component like this. However, my memory is fuzzy, and if I misunderstood, let me know and I'll adjust it accordingly.

ctrueden:
@hadim Feel free to build this component and give it a try. I did not have time to test it at all, although Richard's initial unit tests do pass. If there are problems, please file them as issues in the new repository. Thanks!

rimadoma:
@ctrueden 
OK, mark me down as a developer

PR #249, opened 2016-10-15
Create Initializable

lnyng:
See [imagej-ops#455](https://github.com/imagej/imagej-ops/issues/455)

This interface is moved from imagej-ops. It might be easier for the
underlying class inside the module to implement this interface than to
specify the initializer inside the Plugin annotation, especially because
annotation could not be partially overridden. Now the Initializable
interface has higher priority than the initializer annotation, and at
most one of them will be called when a module is initialized.

### DIFF
--- a/src/main/java/org/scijava/Initializable.java
+++ b/src/main/java/org/scijava/Initializable.java
@@ -1,0 +1,45 @@
+/*
+ * #%L
+ * ImageJ software for multidimensional image processing and analysis.
+ * %%
+ * Copyright (C) 2014 - 2016 Board of Regents of the University of
+ * Wisconsin-Madison, University of Konstanz and Brian Northan.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ * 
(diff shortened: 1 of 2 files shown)


src/main/java/org/scijava/module/AbstractModule.java
```diff
@@ -76,11 +78,16 @@ public void cancel() {
 	public void initialize() throws MethodCallException {
 		// execute global module initializer
 		final Object delegateObject = getDelegateObject();
-		if (initializerRef == null) {
-			final String initializer = getInfo().getInitializer();
-			initializerRef = new MethodRef(delegateObject.getClass(), initializer);
+		if (delegateObject instanceof Initializable) {
+			((Initializable) delegateObject).initialize();
+		}
+		else {
```
ctrueden: Do we need to do either/or here? If so, we should issue a warning when the module implements `Initializable` _and_ declares an initializer method. Or simpler: why not just run both initializers if both are available/specified?

src/main/java/org/scijava/Initializable.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * #%L
+ * ImageJ software for multidimensional image processing and analysis.
+ * %%
+ * Copyright (C) 2014 - 2016 Board of Regents of the University of
```
ctrueden: This copyright header should be updated to match the rest of SJC.

src/main/java/org/scijava/module/AbstractModule.java
ctrueden: There is also a question of handling the case where the `Module` itself is `Initializable`. This can happen e.g. for modules which extend `AbstractModule` directly. I think we should check that instead, and then have `CommandModule` implement `Initializable`, working analogously to [how support works for `Cancelable`](https://github.com/scijava/scijava-common/blob/scijava-common-2.60.2/src/main/java/org/scijava/command/CommandModule.java#L208-L233).

PR #253, opened 2016-12-21
Revamp the LogService

ctrueden:
As discussed with @fjug in Dresden:

1. The `LogService` now sports a spiffy callback mechanism, so that you can register listeners for log events.
2. You can now create any number of `Logger` channels keyed by name, of which the `LogService` itself functions as the `"default"` one, preserving the convenience of the previous paradigm.
3. The `ConsoleService` now forwards all stdout and stderr messages to special channels of the `LogService` named `"stdout"` and `"stderr"` respectively.

@fjug Please review and test with your logging UI. I will merge once you sign off on it.

### DIFF
--- a/src/main/java/org/scijava/console/DefaultConsoleService.java
+++ b/src/main/java/org/scijava/console/DefaultConsoleService.java
@@ -39,7 +39,9 @@ import java.util.List;
 
 import org.scijava.Context;
 import org.scijava.console.OutputEvent.Source;
+import org.scijava.log.LogLevel;
 import org.scijava.log.LogService;
+import org.scijava.log.Logger;
 import org.scijava.plugin.AbstractHandlerService;
 import org.scijava.plugin.Parameter;
 import org.scijava.plugin.Plugin;
@@ -58,11 +60,17 @@ public class DefaultConsoleService extends
 	ConsoleService
 {
 
+	private static final String STDOUT_LOG_CHANNEL = "stdout";
+	private static final String STDERR_LOG_CHANNEL = "stderr";
+
 	@Parameter
 	private ThreadService threadService;
 
 	@Parameter
 	private LogService log;
+
+	private Logger stdoutLogger;
+	private Logger stderrLogger;
 
 	private MultiPrintStream sysout, syserr;
 	private OutputStreamReporter out, err;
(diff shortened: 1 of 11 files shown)


ctrueden:
One thing I did not do, and which needs testing: We want the ability to filter stdout and stderr console events based on the producer (i.e., generating class/package). We need to inspect the stack trace to figure that out. The LogService already has this feature, so probably, nothing new needs to be implemented. We can simply use the stdout and stderr log channels. However, the `callingClass()` method of AbstractLogService almost certainly needs tweaking. We want to ignore anything in org.scijava.log and org.scijava.console completely. But we also don't want the org.scijava.log package to reference org.scijava.console at all. Consider how best to achieve this...

ctrueden:
Remaining action items:

  * [ ] @fjug try the new API. Update his Swing UI logging panels to: 1) not use logback explicitly at all; instead: 2) use the above; 3) report back on what works, what doesn't, etc.
  * [ ] Think about how to fix `org.scijava.ui.console` package to instead be `org.scijava.ui.log`. Most of the design will probably be very similar: `OutputEvent` ➡️ `LogEvent`.
  * [ ] Iterate on the design.
  * [ ] Once it's ready: migrate Swing UI logging panels into `scijava-ui-swing` for everyone's profit.

maarzt:
Hey @ctrueden, I'm sorry for starting a fork of this branch before contributing to this discussion. It was simply the workflow I used for the plot-service branch. But discussions are important so let's start:

maarzt:
At first, I think we should clarify some user stories this branch is dedicated to:

1. User story: While using ImageJ / Fiji log messages should be easy accessible in a central logging window. The user should no be disturbed by multiple log, console and third party log windows popping up.

2. User story: Plugin A uses the functionality of plugin B. Plugin A does not want plugin B to open a logging window. But plugin A still wants plugin B's log messages to be visible in a central logging window or in the logging panel of plugin A.

maarzt:
What does not work:
* I cannot write a logging window that listens to all logging channels. Currently this requires the logging window to register for all logging channels. The problem: This process needs to be repeated again and again, as new logging channels might be added. I see two possible solutions: 1) Having a notification when new channels are added. 2) Providing a listener interface, that is fed with the log messages of all channels (This needs an additional parameter, to figure out, which channel a log message belongs to.)

* I think we shouldn't forward stdout and stderr to logging channels. Logs and streams are different concepts: logs are organized in messages, streams are not. Output to streams is pre formatted and needs a mono spaced font to be displayed correctly. Logs are not pre formatted and can be displayed with any font.
I think two different Listener interfaces for two different concepts was already very good: Listen to log messages by registering to LogService, and listen to stdout/err by registering to ConsoleService. 

* User story: I want all channels, log messages to be forwarded to stdout/stderr as I want to see them in my terminal or IDE.  But I don't want my plugin that listens to the ConsoleService to get the log messages, as it already gets these messages by listening to LogService.
Currently there is no way to write log messages to stdout/stderr while not forwarding them to the listener of ConsoleService. Currently only the messages of the default channel are forwarded to stdout/stderr.

* It would be useful to have an upper limit for log levels. This would simplify algorithms that process log messages depending on their log channel. We should consider making LogLevel an enum.

ctrueden:
Closing in favor of #255.

PR #255, opened 2017-01-27
Log revamp

maarzt:
Proposals concerning pull request https://github.com/scijava/scijava-common/pull/253:
* Don't publish stdout/stderr to log channels, log messages and stream are too different in their concepts.
* Refactoring: Use CopyOnWriteArrayList to manage LogListeners, extract LogFormatter from StderrLogService, make LogLevel an enum.
* Make DefaultLogService::channel(String name) thread safe
* Add the ability to listen to all log channels at once. Use this in StderrLogService.

### DIFF
--- a/src/main/java/org/scijava/console/ConsoleService.java
+++ b/src/main/java/org/scijava/console/ConsoleService.java
@@ -31,6 +31,7 @@
 
 package org.scijava.console;
 
+import java.io.PrintStream;
 import java.util.LinkedList;
 
 import org.scijava.plugin.HandlerService;
--- a/src/main/java/org/scijava/console/DefaultConsoleService.java
+++ b/src/main/java/org/scijava/console/DefaultConsoleService.java
@@ -36,6 +36,7 @@ import java.io.PrintStream;
 import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 import org.scijava.Context;
 import org.scijava.console.OutputEvent.Source;
@@ -68,9 +69,7 @@ public class DefaultConsoleService extends
 	private OutputStreamReporter out, err;
 
 	/** List of listeners for {@code stdout} and {@code stderr} output. */
-	private ArrayList<OutputListener> listeners;
-
-	private OutputListener[] cachedListeners;
+	private List<OutputListener> listeners;
 
 	// -- ConsoleService methods --
 
@@ -115,26 +114,19 @@ public class DefaultConsoleService extends
 	@Override
 	public void addOutputListener(final OutputListener l) {
 		if (listeners == null) initListeners();
-		synchronized (listeners) {
-			listeners.add(l);
-			cacheListeners();
-		}
+		listeners.add(l);
 	}
 
 	@Override
 	public void removeOutputListener(final OutputListener l) {
 		if (listeners == null) initListeners();
-		synchronized (listeners) {
-			listeners.remove(l);
-			cacheListeners();
-		}
+		listeners.remove(l);
 	}
 
 	@Override
 	public void notifyListeners(final OutputEvent event) {
 		if (listeners == null) initListeners();
-		final OutputListener[] toNotify = cachedListeners;
-		for (final OutputListener l : toNotify)
+		for (final OutputListener l : listeners)
 			l.outputOccurred(event);
 	}
 
@@ -162,15 +154,10 @@ public class DefaultConsoleService extends
 		err = new OutputStreamReporter(Source.STDERR);
 		syserr.getParent().addOutputStream(err);
 
-		listeners = new ArrayList<>();
-		cachedListeners = listeners.toArray(new OutputListener[0]);
+		listeners = new CopyOnWriteArrayList<>();
 	}
 
 	// -- Helper methods --
-
-	private void cacheListeners() {
-		cachedListeners = listeners.toArray(new OutputListener[listeners.size()]);
-	}
 
 	private MultiPrintStream multiPrintStream(final PrintStream ps) {
 		if (ps instanceof MultiPrintStream) return (MultiPrintStream) ps;
--- a/src/main/java/org/scijava/console/DefaultConsoleService.java
+++ b/src/main/java/org/scijava/console/DefaultConsoleService.java
@@ -32,13 +32,16 @@
 package org.scijava.console;
 
 import java.io.OutputStream;
+import java.io.PrintStream;
 import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.function.BiFunction;
 
 import org.scijava.Context;
 import org.scijava.console.OutputEvent.Source;
+import org.scijava.log.LogLevel;
 import org.scijava.log.LogService;
 import org.scijava.plugin.AbstractHandlerService;
 import org.scijava.plugin.Parameter;
@@ -70,6 +73,13 @@ public class DefaultConsoleService extends
 	private List<OutputListener> listeners;
 
 	// -- ConsoleService methods --
+
+	@Override
+	public void initialize() {
+		PrintStream logErr = logStream(Source.STDERR);
+		PrintStream logOut = logStream(Source.STDOUT);
+		log.setPrintStreams(level -> (level <= LogLevel.WARN) ? logErr : logOut);
+	}
 
 	@Override
 	public void processArgs(final String... args) {
@@ -128,6 +138,27 @@ public class DefaultConsoleService extends
 			l.outputOccurred(event);
 	}
 
+	PrintStream logStream(Source source) {
+		OutputStream a = getBypassingStream(source);
+		OutputStream b = new OutputStreamReporter((relevance, text) -> {
+			final Context context = getContext();
+			final boolean contextual = true;
+			final boolean containsLog = true;
+			return new OutputEvent(context, source, text, contextual, containsLog);
+		});
+		return new PrintStream(new MultiOutputStream(a, b));
+	}
+
+	private OutputStream getBypassingStream(Source source) {
+		switch (source) {
+			case STDOUT:
+				return ListenableSystemStreams.out().bypass();
+			case STDERR:
+				return ListenableSystemStreams.err().bypass();
+		}
+		throw new AssertionError();
+	}
+
 	// -- Disposable methods --
 
 	@Override
@@ -142,12 +173,21 @@ public class DefaultConsoleService extends
 	private synchronized void initListeners() {
 		if (listeners != null) return; // already initialized
 
-		out = new OutputStreamReporter(Source.STDOUT);
+		out = setupDefaultReporter(Source.STDOUT);
+		err = setupDefaultReporter(Source.STDERR);
 		ListenableSystemStreams.out().addOutputStream(out);
-		err = new OutputStreamReporter(Source.STDERR);
 		ListenableSystemStreams.err().addOutputStream(err);
 
 		listeners = new CopyOnWriteArrayList<>();
+	}
+
+	private OutputStreamReporter setupDefaultReporter(Source source) {
+		return new OutputStreamReporter((relevance, output) -> {
+			final Context context = getContext();
+			final boolean contextual = relevance == ThreadContext.SAME;
+			final boolean containsLog = false;
+			return new OutputEvent(context, source, output, contextual, containsLog);
+		});
 	}
 
 	// -- Helper methods --
@@ -177,11 +217,12 @@ public class DefaultConsoleService extends
 	 */
 	private class OutputStreamReporter extends OutputStream {
 
-		/** Source of the output stream; i.e., {@code stdout} or {@code stderr}. */
-		private final Source source;
+		private final BiFunction<ThreadContext, String, OutputEvent> eventFactory;
 
-		public OutputStreamReporter(final Source source) {
-			this.source = source;
+		public OutputStreamReporter(
+			BiFunction<ThreadContext, String, OutputEvent> eventFactory)
+		{
+			this.eventFactory = eventFactory;
 		}
 
 		// -- OutputStream methods --
@@ -207,12 +248,8 @@ public class DefaultConsoleService extends
 		}
 
 		private void publish(final ThreadContext relevance, final String output) {
-			final Context context = getContext();
-			final boolean contextual = relevance == ThreadContext.SAME;
-			final OutputEvent event =
-				new OutputEvent(context, source, output, contextual);
+			final OutputEvent event = eventFactory.apply(relevance, output);
 			notifyListeners(event);
 		}
 	}
-
 }
--- a/src/main/java/org/scijava/console/DefaultConsoleService.java
+++ b/src/main/java/org/scijava/console/DefaultConsoleService.java
@@ -32,7 +32,6 @@
 package org.scijava.console;
 
 import java.io.OutputStream;
-import java.io.PrintStream;
 import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
@@ -65,7 +64,6 @@ public class DefaultConsoleService extends
 	@Parameter
 	private LogService log;
 
-	private MultiPrintStream sysout, syserr;
 	private OutputStreamReporter out, err;
 
 	/** List of listeners for {@code stdout} and {@code stderr} output. */
@@ -134,8 +132,8 @@ public class DefaultConsoleService extends
 
 	@Override
 	public void dispose() {
-		if (out != null) sysout.getParent().removeOutputStream(out);
-		if (err != null) syserr.getParent().removeOutputStream(err);
+		if(out != null) ListenableSystemStreams.out().removeOutputStream(out);
+		if(err != null) ListenableSystemStreams.err().removeOutputStream(err);
 	}
 
 	// -- Helper methods - lazy initialization --
@@ -144,25 +142,15 @@ public class DefaultConsoleService extends
 	private synchronized void initListeners() {
 		if (listeners != null) return; // already initialized
 
-		sysout = multiPrintStream(System.out);
-		if (System.out != sysout) System.setOut(sysout);
 		out = new OutputStreamReporter(Source.STDOUT);
-		sysout.getParent().addOutputStream(out);
-
-		syserr = multiPrintStream(System.err);
-		if (System.err != syserr) System.setErr(syserr);
+		ListenableSystemStreams.out().addOutputStream(out);
 		err = new OutputStreamReporter(Source.STDERR);
-		syserr.getParent().addOutputStream(err);
+		ListenableSystemStreams.err().addOutputStream(err);
 
 		listeners = new CopyOnWriteArrayList<>();
 	}
 
 	// -- Helper methods --
-
-	private MultiPrintStream multiPrintStream(final PrintStream ps) {
-		if (ps instanceof MultiPrintStream) return (MultiPrintStream) ps;
-		return new MultiPrintStream(ps);
-	}
 
 	/**
 	 * Gets whether two lists have exactly the same elements in them.
--- a/src/main/java/org/scijava/log/AbstractLogService.java
+++ b/src/main/java/org/scijava/log/AbstractLogService.java
@@ -54,8 +54,8 @@ public abstract class AbstractLogService extends AbstractService implements
 	// -- constructor --
 
 	public AbstractLogService() {
-		logger = new DefaultListenableLogger(this::messageLogged, LogSource
-			.root(), LogLevel.NONE)
+		logger = new LogServiceLogger(this::messageLogged, LogSource.root(),
+			LogLevel.NONE)
 		{
 
 			@Override
@@ -74,7 +74,12 @@ public abstract class AbstractLogService extends AbstractService implements
 
 	@Override
 	public void setLevel(String classOrPackageName, final int level) {
-		logLevelStrategy.setLevel(classOrPackageName, level);
+		logLevelStrategy.setLevelForClass(classOrPackageName, level);
+	}
+
+	@Override
+	public void setLevelForLogger(LogSource source, final int level) {
+		logLevelStrategy.setLevelForLogger(source, level);
 	}
 
 	abstract void messageLogged(LogMessage message);
@@ -121,5 +126,27 @@ public abstract class AbstractLogService extends AbstractService implements
 	@Override
 	public ListenableLogger listenableLogger(int level) {
 		return logger.listenableLogger(level);
+	}
+
+	@IgnoreAsCallingClass
+	private class LogServiceLogger extends DefaultListenableLogger {
+
+		public LogServiceLogger(LogListener destination, LogSource source,
+			int level)
+		{
+			super(destination, source, level);
+		}
+
+		@Override
+		public ListenableLogger subLogger(String name, int level) {
+			LogSource source = getSource().subSource(name);
+			int actualLevel = logLevelStrategy.getLevelForLogger(source, level);
+			return new LogServiceLogger(this, source, actualLevel);
+		}
+
+		@Override
+		public ListenableLogger listenableLogger(int level) {
+			return new LogServiceLogger(this, getSource(), level);
+		}
 	}
 }
--- a/src/main/java/org/scijava/log/AbstractLogService.java
+++ b/src/main/java/org/scijava/log/AbstractLogService.java
@@ -114,102 +114,102 @@ public abstract class AbstractLogService extends AbstractService implements
 
 	@Override
 	public void debug(final Object msg) {
-		log(DEBUG, msg, null);
+		log(LogLevel.DEBUG, msg, null);
 	}
 
 	@Override
 	public void debug(final Throwable t) {
-		log(DEBUG, null, t);
+		log(LogLevel.DEBUG, null, t);
 	}
 
 	@Override
 	public void debug(final Object msg, final Throwable t) {
-		log(DEBUG, msg, t);
+		log(LogLevel.DEBUG, msg, t);
 	}
 
 	@Override
 	public void error(final Object msg) {
-		log(ERROR, msg, null);
+		log(LogLevel.ERROR, msg, null);
 	}
 
 	@Override
 	public void error(final Throwable t) {
-		log(ERROR, null, t);
+		log(LogLevel.ERROR, null, t);
 	}
 
 	@Override
 	public void error(final Object msg, final Throwable t) {
-		log(ERROR, msg, t);
+		log(LogLevel.ERROR, msg, t);
 	}
 
 	@Override
 	public void info(final Object msg) {
-		log(INFO, msg, null);
+		log(LogLevel.INFO, msg, null);
 	}
 
 	@Override
 	public void info(final Throwable t) {
-		log(INFO, null, t);
+		log(LogLevel.INFO, null, t);
 	}
 
 	@Override
 	public void info(final Object msg, final Throwable t) {
-		log(INFO, msg, t);
+		log(LogLevel.INFO, msg, t);
 	}
 
 	@Override
 	public void trace(final Object msg) {
-		log(TRACE, msg, null);
+		log(LogLevel.TRACE, msg, null);
 	}
 
 	@Override
 	public void trace(final Throwable t) {
-		log(TRACE, null, t);
+		log(LogLevel.TRACE, null, t);
 	}
 
 	@Override
 	public void trace(final Object msg, final Throwable t) {
-		log(TRACE, msg, t);
+		log(LogLevel.TRACE, msg, t);
 	}
 
 	@Override
 	public void warn(final Object msg) {
-		log(WARN, msg, null);
+		log(LogLevel.WARN, msg, null);
 	}
 
 	@Override
 	public void warn(final Throwable t) {
-		log(WARN, null, t);
+		log(LogLevel.WARN, null, t);
 	}
 
 	@Override
 	public void warn(final Object msg, final Throwable t) {
-		log(WARN, msg, t);
+		log(LogLevel.WARN, msg, t);
 	}
 
 	@Override
 	public boolean isDebug() {
-		return getLevel() >= DEBUG;
+		return getLevel() >= LogLevel.DEBUG;
 	}
 
 	@Override
 	public boolean isError() {
-		return getLevel() >= ERROR;
+		return getLevel() >= LogLevel.ERROR;
 	}
 
 	@Override
 	public boolean isInfo() {
-		return getLevel() >= INFO;
+		return getLevel() >= LogLevel.INFO;
 	}
 
 	@Override
 	public boolean isTrace() {
-		return getLevel() >= TRACE;
+		return getLevel() >= LogLevel.TRACE;
 	}
 
 	@Override
 	public boolean isWarn() {
-		return getLevel() >= WARN;
+		return getLevel() >= LogLevel.WARN;
 	}
 
 	@Override
@@ -257,7 +257,7 @@ public abstract class AbstractLogService extends AbstractService implements
 
 	private int levelFromEnvironment() {
 		// use the default, which is INFO unless the DEBUG env. variable is set
-		return System.getenv("DEBUG") == null ? INFO : DEBUG;
+		return System.getenv("DEBUG") == null ? LogLevel.INFO : LogLevel.DEBUG;
 	}
 
 }
--- a/src/main/java/org/scijava/log/AbstractLogService.java
+++ b/src/main/java/org/scijava/log/AbstractLogService.java
@@ -46,7 +46,7 @@ public abstract class AbstractLogService extends AbstractService implements
 	LogService
 {
 
-	private int currentLevel = System.getenv("DEBUG") == null ? INFO : DEBUG;
+	private int currentLevel = levelFromEnvironment();
 
 	private final Map<String, Integer> classAndPackageLevels =
 		new HashMap<>();
@@ -77,10 +77,8 @@ public abstract class AbstractLogService extends AbstractService implements
 		final int level = level(logProp);
 		if (level >= 0) setLevel(level);
 
-		if (getLevel() == 0) {
-			// use the default, which is INFO unless the DEBUG env. variable is set
-			setLevel(System.getenv("DEBUG") == null ? INFO : DEBUG);
-		}
+		if (getLevel() == 0)
+			setLevel(levelFromEnvironment());
 
 		// populate custom class- and package-specific log level properties
 		final String logLevelPrefix = LOG_LEVEL_PROPERTY + ":";
@@ -295,6 +293,11 @@ public abstract class AbstractLogService extends AbstractService implements
 		final int dot = classOrPackageName.lastIndexOf(".");
 		if (dot < 0) return null;
 		return classOrPackageName.substring(0, dot);
+	}
+
+	private int levelFromEnvironment() {
+		// use the default, which is INFO unless the DEBUG env. variable is set
+		return System.getenv("DEBUG") == null ? INFO : DEBUG;
 	}
 
 }
--- a/src/main/java/org/scijava/log/AbstractLogService.java
+++ b/src/main/java/org/scijava/log/AbstractLogService.java
@@ -54,7 +54,9 @@ public abstract class AbstractLogService extends AbstractService implements
 	// -- constructor --
 
 	public AbstractLogService() {
-		logger = new DefaultListenableLogger(this::messageLogged, "", LogLevel.NONE) {
+		logger = new DefaultListenableLogger(this::messageLogged, LogSource
+			.root(), LogLevel.NONE)
+		{
 
 			@Override
 			public int getLevel() {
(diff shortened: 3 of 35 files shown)


ctrueden:
Hey @maarzt, just wanted to chime in quickly that I have not forgotten about this work! It is high up on my priority list now. I probably will not have time to check it today, but will try to go through everything on Monday.

maarzt:
Hi @ctrueden,
that's great. I used the new api in @fjug's Tr2d plugin, but sadly, it didn't worked out very well. There are still open questions not answered by the current design. And I don't feel very comfortable with making design changes without discussion.
* How do we want to set the log level, now that there are multiple log channels? Who is responsible/able to set the Level?
* In Tr2d many log messages are lost, before everything is fully set up. A coherent concept, how log messages are delivered to it's destination and how to set up loggers, seems to be necessary.
* You wanted to forward system.out and err to log channels, to get a context for them. What would you like to use this information for?
Sorry for the many questions... Matthias

ctrueden:
Quick update: I still, frustratingly, have not gotten to this code review + discussion. It is still at/near the very top of my list. Unfortunately, I first need to fix the broken HTTPS certs on one of the SciJava server machines (so that e.g. https://maven.imagej.net/ starts working again, and continues working into the future, including automatic cert renewals). There is no way I will complete that work, and circle back around to this PR, before the end of this week. However, I just wanted to let you know that it is still very much on my radar, in queue.

maarzt:
Hey @ctrueden, 

I thought about this logger stuff again and again. Mainly I could not find any useful application of the  LogService.channel feature introduced in this PR. (Since I don't like forwarding stdout and err to log channels.)

After some discussions, I think the current solution in master is quite close to a pretty good solution. When executing a Command, we give an object to it (LogService), which is used by to Command for logging. This pattern is simple and powerful. As the caller, who provides LogService, has full control of the log messages.

This gives a simple solution to the user story I described in https://github.com/scijava/scijava-common/pull/253:
2. User story: Plugin A uses the functionality of plugin B. Plugin A does not want plugin B to open a logging window. But plugin A still wants plugin B's log messages to be visible in a central logging window or in the logging panel of plugin A.

Plugin A simply creates it's custom LogService and hands it over the plugin B when calling pulgin B. This gives plugin A full control over plugin B's log messages. Plugin A can display the messages in it's on logging panel, and it can forward the logging messages to a central logging window.

Problem solved without changing scijav. Cheers

maarzt:
But there is a second goal. **We want the Console window to be smarter when dealing with log messages.** First of all the Console window should listen to ConsoleService to get stdout and stderr. Second the Console window should listen directly to LogService to get the all log messages with all meta data. Then I will add functionality to the Console window like search, filtering, log level adjustment and customizable pop up conditions, and it will be much smarter.

Here comes the third goal. **We want meta data, that tells us where the log messages come from.** But how do we retrieve this information. I'm afraid inspection the stack trace will be slow an buggy. Because retrieving the stack trace is slow (throwing a exception is slightly dirty hack). And the question: Which class to pick from the call stack is hard to answer, and will result in unexpected behavior.
I would suggest a different solution, which is against the rules, because I would like to have several "LogServices". **One LogService instance per Command that is executed.** Each LogService knows which Command it belongs to, and it append's this information to the log messages it get's. Then the log messages are forwarded to a central log service like thing.

@ctrueden What do you think. Might this idea be better than inspecting the stack trace? What would be the way to go, without breaking the rules.

src/main/java/org/scijava/log/DefaultLogger.java
```diff
@@ -31,10 +31,8 @@
 
 package org.scijava.log;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Properties;
+import java.util.*;
```
ctrueden: SciJava style is to never use wildcard imports.

And in general: please keep changes in a patch set atomic, and matching what the commit message says is being changed. If you really wanted to change the imports like this, it should be its own commit.

src/main/java/org/scijava/log/DefaultLogger.java
```diff
@@ -49,10 +47,7 @@
 
 	private String name;
 
-	/** List of listeners for logging events. */
```
ctrueden: Why delete the javadoc? It wasn't hurting anything.

src/main/java/org/scijava/log/DefaultLogger.java
```diff
@@ -49,10 +47,7 @@
...
-	private ArrayList<LogListener> listeners;
-
-	private LogListener[] cachedListeners;
+	private List<LogListener> listeners = new CopyOnWriteArrayList<>();
```
ctrueden: This is very cool to know about. Thanks!

src/main/java/org/scijava/log/DefaultLogger.java
```diff
@@ -116,29 +111,19 @@ public void alwaysLog(final int level, final Object msg, final Throwable t) {
 
 	@Override
 	public void addLogListener(final LogListener l) {
-		if (listeners == null) initListeners();
-		synchronized (listeners) {
-			listeners.add(l);
-			cacheListeners();
-		}
+		listeners.add(Objects.requireNonNull(l));
```
ctrueden: I agree that we should be requiring non-null here at this level. However, that is a separate conceptual change, which would be better as its own commit. Or else the commit message here should state that it is also changing that behavior.

src/test/java/org/scijava/log/MultiThreadingTest.java
```diff
@@ -0,0 +1,78 @@
+package org.scijava.log;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.*;
```
ctrueden: Please use the IDE's import organization function to enumerate all the classes, instead of using wildcards.

src/test/java/org/scijava/log/MultiThreadingTest.java
```diff
@@ -0,0 +1,78 @@
...
+	private static <T> List<T> runParallelCollectResults(int nTasks, Callable<T> callable) throws InterruptedException {
+		ExecutorService executorService = Executors.newFixedThreadPool(nTasks);
+		List<Future<T>> futures = executorService.invokeAll(Collections.nCopies(nTasks, callable));
+		return futures.stream().map(MultiThreadingTest::removeFuture).collect(Collectors.toList());
+	}
+
+	private static <T> T removeFuture(Future<T> future) {
+		try {
+			return future.get();
+		} catch (InterruptedException | ExecutionException e) {
```
ctrueden: SciJava style is:
```java
try {
	doStuff();
}
catch (final MyException exc) {
	handleStuff();
}
```

src/test/java/org/scijava/log/MultiThreadingTest.java
```diff
@@ -0,0 +1,78 @@
...
+			return null;
+		}
+	}
+
+	private static boolean referencesInListEqual(List<? extends Object> a, List<? extends Object> b) {
+		Iterator<?> ia = a.iterator();
+		Iterator<?> ib = b.iterator();
+		boolean result = a.size() == b.size();
+		while(result && ia.hasNext() && ib.hasNext())
```
ctrueden: SciJava style is `while (condition)` with leading space.

src/test/java/org/scijava/log/MultiThreadingTest.java
```diff
@@ -0,0 +1,78 @@
...
+			result = ia.next() == ib.next();
+		return result;
+	}
+
+	private static <T> boolean allEqual(Iterable<T> l, BiFunction<T, T, Boolean> equals) {
```
ctrueden: Super general! Shall we put this into `org.scijava.test.TestUtils`?

src/test/java/org/scijava/log/MultiThreadingTest.java
```diff
@@ -0,0 +1,78 @@
...
+	@Before
+	public void setUp() {
+		logService = new StderrLogService();
+	}
+
+	@Test
+	public void run() throws InterruptedException {
+		List<List<Logger>> results = runParallelCollectResults(10, this::getLoggers);
+		boolean passed = allEqual(results, MultiThreadingTest::referencesInListEqual);
+		assertTrue(passed);
```
ctrueden: Could collapse this into an `assertAllEqual` static method in `org.scijava.test.TestUtils`. Of course, there is value in having `allEqual` which returns `boolean` as well. How about if that lives in `org.scijava.util` somewhere? Right now we have `ArrayUtils` and `ListUtils`. However, we are moving more toward the ImgLib2-style (and core Java style) of plurals for utility classes. Any suggestions? Happy to move/reorganize/deprecate existing stuff into better structure along these lines.

src/main/java/org/scijava/log/AbstractLogService.java
```diff
@@ -107,7 +125,7 @@ public void removeLogListener(final LogListener l) {
 	}
 
 	@Override
-	public void notifyListeners(final int level, final Object msg,
+	public void notifyListeners(final LogLevel level, final Object msg,
```
ctrueden: > The Integer enum pattern has a lot of drawbacks.

So, in general I very much agree with you. However, in this case, I think using integers, or maybe even doubles, actually makes more sense, for similar reasons to why we use doubles for priority (see `org.scijava.Priority`).

> What means level == 0 or LogLevel.NONE? (It's possible meaning "unknown log level" is better expressed by a null reference.)

The idea of 0 i.e. `NONE` is that _nothing_ should be logged, of course. That is different than "unknown" or null. Actually, I would not allow nulls here... another reason `int` happens to be handy, since primitives cannot be null.

> What means level == -1? (Better throw an exception if it's an error.)

I agree that negative levels are not defined here, since 0/NONE is the bottommost behavior. So, we should indeed perform bounds checking and throw an exception if negative values are given anywhere in this API.

> What's the maximum log level? (Now it's clearly LogLevel.TRACE.)

Does there need to be an explicit maximum? I don't think so. It is always possible to be "more verbose" than everything else that has ever existed...

src/main/java/org/scijava/log/LogLevel.java
```diff
@@ -31,66 +31,72 @@
...
+			case 2: return WARN;
+			case 3: return INFO;
+			case 4: return DEBUG;
+			case 5: return TRACE;
 		}
-		return -1;
+		throw new IllegalArgumentException();
+	}
+
+	public boolean isHigherThan(LogLevel level) {
```
ctrueden: Why not implement `Comparable`? Although again: I personally do not think we need/want an enum here. Too restrictive. And an extensible enum pattern seems like overkill, since there is no additional metadata we need to attach to each level, besides its `int` value.

ctrueden:
Some general comments about the commit message style: we use imperative tense, with no period in the subject line. Please avoid prefixes like "Refactor:" and "Refactoring:". See the [Coding style page](http://imagej.net/Coding_style#SCM_history), which links to [this](https://chris.beams.io/posts/git-commit/) and [that](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) articles on Git commit messages.

src/main/java/org/scijava/log/DefaultLogFormatter.java
```diff
@@ -0,0 +1,57 @@
+/*
```
ctrueden: Please remove the conflicts list from the commit message. It's not relevant to the actual patch, only some transient rebase.

src/main/java/org/scijava/log/DefaultLogFormatter.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+package org.scijava.log;
+
+import org.scijava.Named;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+
+/**
+ * @author Matthias Arzt
```
ctrueden: Please add at least a one-sentence description of the class to the javadoc here.

src/main/java/org/scijava/log/DefaultLogFormatter.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+public class DefaultLogFormatter implements LogFormatter{
```
ctrueden: SciJava style is `implements FooBar {` with leading space before the curly brace.

src/test/java/org/scijava/log/DefaultLogFormatterTest.java
```diff
@@ -0,0 +1,107 @@
...
+ */
+
+package org.scijava.log;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.scijava.Named;
+
+/**
+ * @author Matthias Arzt
```
ctrueden: Please add javadoc. The easiest and most standards for tests is simply:
```java
/**
 * Tests {@link DefaultLogFormatter}.
 *
 * @author Matthias Arzgt
 */
```

src/test/java/org/scijava/log/DefaultLogFormatterTest.java
```diff
@@ -0,0 +1,107 @@
...
+		Throwable t;
+		try {
+			throw new NullPointerException();
+		} catch (Throwable e) {
+			t = e;
+		}
+		return t;
+	}
+
+	class NamedObject implements Named {
```
ctrueden: Why not `private class`, just to be safe?

src/main/java/org/scijava/log/LogLevel.java
ctrueden: Actually, there is one single piece of metadata: the label. E.g., 0=NONE, 3=INFO. This would be nice if it were extensible... and actually, it's easy to do so by following the extensible enum pattern. Take a look at `net.imagej.axis.AxisType` and `net.imagej.axis.Axes`. Would this sort of pattern satisfy you here, @maarzt? It is significantly more complex, but would allow custom log levels to name themselves. Otherwise, we could achieve the same thing by adding a method to `LogService` which lets people register new labels with custom log levels. This seems overly fancy and potentially error-prone to me, since we need to implement lots of methods to support it. E.g.: what if two `LogLevel`s exist with the same `int` level but different labels? Easier to simply stick with `int`.

src/main/java/org/scijava/log/DefaultLogFormatter.java
```diff
@@ -0,0 +1,57 @@
...
+
+	@Override
+	public String format(Named source, LogLevel level, Object msg, Throwable t) {
+		final StringWriter sw = new StringWriter();
+		final PrintWriter message = new PrintWriter(sw);
+		message.print(level.prefix());
+		final String s = source.getName();
+		if (!s.equals("default"))
```
ctrueden: This seems really hacky, to do a string check for the name `"default"` instead of something based on Java types.

src/main/java/org/scijava/log/StderrLogService.java
```diff
@@ -49,18 +51,12 @@
 @Plugin(type = Service.class, priority = Priority.LOW_PRIORITY)
 public class StderrLogService extends AbstractLogService {
 
+	private final LogFormatter formatter = new DefaultLogFormatter();
+
 	@Override
 	public void alwaysLog(final LogLevel level, final Object msg, final Throwable t) {
-		final String message = level.prefix() + msg;
-		// NB: Emit severe messages to stderr, and less severe ones to stdout.
```
ctrueden: I appreciate refactoring to make things easier to read and understand, but it is very important to keep the `NB` notes instead of throwing them away. Every `NB` note indicates something of particular note, and often unintuitive, to developers reading the code. See http://imagej.net/Coding_style#Javadoc_and_comments

src/main/java/org/scijava/log/AbstractLogService.java
ctrueden: And there is another reason not to introduce `LogLevel`: it breaks backwards API compatibility. The changes done earlier in this topic branch were done very carefully so that existing consumers of `LogService` will actually keep working without recompilation—at least in theory. :wink: If you want to introduce `LogLevel`, we need to keep _all_ the existing signatures that use `int level` also, deprecate them, and have them delegate to the new signatures. Otherwise, this change needs to target SJC 3.0.0, which is the next major i.e. breaking release.

src/main/java/org/scijava/log/DefaultLogFormatter.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * @author Matthias Arzt
+ */
+public class DefaultLogFormatter implements LogFormatter{
+
+	@Override
+	public String format(Named source, LogLevel level, Object msg, Throwable t) {
+		final StringWriter sw = new StringWriter();
+		final PrintWriter message = new PrintWriter(sw);
+		message.print(level.prefix());
```
ctrueden: Assuming we stick with the `LogFormatter` idea here, we could then perhaps eliminate `level.prefix()`, since it only exists to facilitate logging, and instead put the brackets etc. right here in the `format` method. What do you think?

src/main/java/org/scijava/console/ConsoleService.java
```diff
@@ -62,6 +63,12 @@
 	/** Notifies listeners of output sent to {@code stdout} or {@code stderr}. */
 	void notifyListeners(OutputEvent event);
 
+	/** Stream that writes to stdout without notifying listeners. */
+	PrintStream getSystemOutBypassListeners();
```
ctrueden: Do you actively and acutely need this? Otherwise, I have serious reservations about including and maintaining this sort of short-circuiting.

src/main/java/org/scijava/console/DefaultConsoleService.java
```diff
@@ -179,11 +195,23 @@ private void cacheListeners() {
...
-	private MultiPrintStream multiPrintStream(final PrintStream ps) {
+	private static MultiPrintStream multiPrintStream(final PrintStream ps) {
 		if (ps instanceof MultiPrintStream) return (MultiPrintStream) ps;
 		return new MultiPrintStream(ps);
 	}
 
+	private static PrintStream skipMultiPrintStream(final PrintStream ps) {
+		if (! (ps instanceof MultiPrintStream) )
+			return ps;
+		return printStream(((MultiPrintStream) ps).getDestination());
```
ctrueden: It's been awhile since I've studied or thought about this logic, but might this need to be recursive, in case of multiple layers of `MultiPrintStream` wrapping?

src/test/java/org/scijava/log/StderrLogServiceTest.java
```diff
@@ -0,0 +1,93 @@
...
+import java.io.OutputStream;
+import java.io.PrintStream;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * @author Matthias Arzt
+ */
+public class StderrLogServiceTest {
```
ctrueden: Instead of a new `StderrLogServiceTest`, this could go in the existing `LogServiceTest`? I know, it's weird to name the unit test class after an interface. This is how things are consistently named for services now, though—I'd rather either: A) stay consistent; or B) rename all the service test classes accordingly. But given that there is nearly always only one default implementation, I'd rather stick with (A).

src/test/java/org/scijava/log/StderrLogServiceTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+		public RedirectSystemOut() {
+			systemOutBefore = System.out;
+			systemErrBefore = System.err;
+			System.setOut(new PrintStream(out));
+			System.setErr(new PrintStream(err));
+		}
+
+		public void close() {
+			System.setOut(systemOutBefore);
```
ctrueden: This sort of global redirection could conceivably cause problems when running tests multi-threaded. I don't really know how to test it without doing that, though, so I'm not suggesting we change it... just wondering if there is a better way.

src/main/java/org/scijava/log/AbstractLogGenerator.java
```diff
@@ -0,0 +1,60 @@
...
+package org.scijava.log;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+/**
+ * @author Matthias Arzt
+ */
+class AbstractLogGenerator {
```
ctrueden: This approach introduces several classes, and feels somehow too complicated to me.

What if we made log channels hierarchical? Then we could have a root channel which is the parent of all other channels. And if you subscribe to the root channel you would thus receive all updates. The default channel would be a child of root, and other channels created with null parent would implicitly be children of root. But you could also create a channel that is explicitly the child of some other channel.

This gets you everything you want, no? And is simpler?

src/main/java/org/scijava/log/StderrLogService.java
```diff
@@ -51,12 +51,25 @@
...
-	private final LogFormatter formatter = new DefaultLogFormatter();
+	public StderrLogService() {
+		addAllChannelsLogListener(new PrintListener());
+	}
+
+	private static class PrintListener implements LogListener {
+
+		private final LogFormatter formatter = new DefaultLogFormatter();
+
+		private final Logger dummySource = new DefaultLogger(); // FIXME add "Logger source" as parameter to messageLogged
```
ctrueden: Sorry, I don't understand this commit. We do _not_ want to emit _all_ channels to stderr/stdout... only the default channel. This was explicitly part of the design decided upon by @fjug and myself. The idea is that you may want to create some additional channels which do not appear on stdout, but which only your particular code is interested in consuming.

src/main/java/org/scijava/log/DefaultLogger.java
```diff
@@ -64,7 +64,7 @@ public void setLogLevel(final String classOrPackageName, final LogLevel level) {
 
 	@Override
 	public void alwaysLog(final LogLevel level, final Object msg, final Throwable t) {
-		notifyListeners(level, msg, t);
+		notifyListeners(this, level, msg, t);
```
ctrueden: Cool, I think this is a good idea. But we need to maintain backwards compatibility. The old `notifyListeners` method should remain (possibly/probably deprecated), with its source being the default channel.

src/test/java/org/scijava/log/GetCallingClassTest.java
```diff
@@ -0,0 +1,35 @@
+package org.scijava.log;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * @author Matthias Arzt
```
ctrueden: Javadoc.

src/main/java/org/scijava/log/IgnoreAsCallingClass.java
```diff
@@ -0,0 +1,10 @@
+package org.scijava.log;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+/**
+ * @author Matthias Arzt
```
ctrueden: Javadoc.

src/main/java/org/scijava/log/IgnoreAsCallingClass.java
```diff
@@ -0,0 +1,10 @@
...
+ */
+@Retention(RetentionPolicy.RUNTIME)
+public @interface IgnoreAsCallingClass {}
```
ctrueden: Super great idea!

src/main/java/org/scijava/log/Utils.java
```diff
@@ -0,0 +1,33 @@
+package org.scijava.log;
+
+import java.util.Iterator;
+
+/**
+ * @author Matthias Arzt
+ */
+
+@IgnoreAsCallingClass
+public class Utils {
```
ctrueden: Let's please not have a utility class called `Utils`. We can name it more explicitly relating to what sort of utility method(s) it provides. How about... `StackTraces` or `Stacks` or `Threads`? There are some similar things already: `DebugUtils.getMainClassName()`, `DebugUtils.getStackDump()`, and `DebugUtils.getStackTrace(Throwable)`, all of which could go into the same utility class as this new method.

src/main/java/org/scijava/log/IgnoreAsCallingClass.java
ctrueden: Please improve the commit message to describe more explicitly what this commit is all about, though!

ctrueden:
OK, I finished reviewing the commits. You have done a lot of really great improvements. But I also have substantial concerns and disagreements with many parts of it. Feel free to go through my comments, make edits and updates, rebase to stay clean, and then force push.

Before that, though, we of course should discuss your greater concerns you stated recently above. I am working on my response to those concerns now, and will edit this here comment to include all of that, as soon as it is ready.

src/test/java/org/scijava/log/MultiThreadingTest.java
maarzt: I will add this coding conventions to http://imagej.net/Coding_style.
And put the description of how to the Eclipse coding style for IntelliJ to a more prominent position.

src/test/java/org/scijava/log/MultiThreadingTest.java
maarzt: I am not sure if `assertAllEqual` will be used more than once.
There are many class in `org.scijava.util` what would be more appropriate `ListUtils` or `MiscUtils`?

src/main/java/org/scijava/console/ConsoleService.java
maarzt: I think we will need this. I want the smarter log window to be able to show the informations written to System.out and System.err separately from the log messages. I really want the LogService to write the logs to the terminal without spamming the OutputListeners.

maarzt:
@ctrueden Thanks for reviewing this code so carefully, and thanks for all comments. I agree with nearly all the things you wrote.

(About inspecting the stack trace to retrieve the calling class)
> Slow I might concede, although I would first ask whether you actually tested it. Buggy, though, I am not buying—especially with your shiny new @IgnoreAsCallingClass annotation.

Ok, I will do the performance test. By Buggy I mean some surprising behavior. For example, somebody wants to clean up his code by moving the log message generating lines to a separate class. Then this changes the calling class. If later one the log messages are filtered by the calling class information, this might result in the log messages not being visible anymore or to appear in the wrong place. At this point 
this person will be surprised and disappointed and instead of figuring out that adding @IgnoreAsCallingClass might solve his problem, this person most likely reverts the cleaning up.

(I can think of situations like list.forEach(elemtent -> element.doTheLogging()), where it would be necessary but impossible to add @IgnoreAsCallingClass to the List class.)

> Here is an alternate proposal: what if, when you call LogService.channel, at that point it inspects the call stack to determine who is asking for the Logger, and it constructs a new object which remembers the calling class, as well as its associated parent Logger?

This would solve the problems, yes.

>Here is one possible approach along these lines:
>1.   Switch to hierarchical loggers, as I suggested above in my review.

yeah, I like this idea.

>2.  Add Class<?> userClass() and/or Object user() to the Logger interface, which return null by default.

I am not sure what's best Class<?> or Object or simply String.

>3.  Have LogService.channel always return a new child Logger which wraps the "base" Logger associated with that channel, but changes the behavior of the user method to be non-null in one of two ways:
>- Introduce channel(Object user, ...) and/or channel(Class<?> userClass, ...) method(s) where the caller can explicitly give the intended user object or class, which will be stored into the child Logger at construction.
>- For the old signatures where no user is given, only then inspect the call stack to infer the user class from the calling class.

There is no such old signature, which we need to implement. This raises the consequence: We wouldn't actually need to inspect the calling class at all.

I still have one concern about this approach. Does it mean every Command should contain the same code:
```
class ExampleCommand implements Command, Initializable {
	@Parameter
	LogService logService;

	Logger log;

	public void initialize() {
		log = logService.channel(this);
	}

	public void run() {
		...
	}
}
```

instead of simply

```
class ExampleCommand implements Command, Initializable {
	@Parameter
	Logger log;

	public void run() {
		...
	}
}
```

I like the LogProcessor Idea.

Thanks for the feedback. I start getting an idea of how to do this in a clean and sound way!

ctrueden:
> I am not sure what's best Class<?> or Object or simply String.

Yeah, I am not sure either. `String` is nice and flexible. `Object` invite memory leaks, I think. So let's stick with either `String` or `Class`.

> There is no such old signature, which we need to implement. This raises the consequence: We wouldn't actually need to inspect the calling class at all.

Not sure I understand your comment here. By "old signature" I meant `Logger channel(String name)` which is of course a "new" signature with respect to `master`, but `old` compared to the idea of having something like `Logger channel(String name, Class<?> userClass)`. My point is that I think we _do_ want to keep `Logger channel(String name)` and have it inspect the call stack to glean the `userClass`.

> I still have one concern about this approach. Does it mean every Command should contain the same code:

No, your latter example is what I had in mind. I.e., this:
```java
	@Parameter
	private Logger log;

	public void run() {
		log.info("I like turtles.");
	}
}
```

This `log` parameter would be, by default, filled in with a `Logger` child of the default channel, whose user class is the delegate class of the module being executed. In the case of `Command` modules, that would be the command class itself. I think it will be highly unusual, once these changes are in place, for commands to need a `LogService` parameter specifically anymore.

I hope this makes sense, but happy to explain in more detail if needed.

> I start getting an idea of how to do this in a clean and sound way!

Cool, it seems like we are converging on a solution, and that this branch is relatively close to that solution. Here are the steps I see still needing to be done:

* Rebase this branch to eliminate edc368f, e227e4c and 9e2aa2d (and hence also the revert commits 7ffe351 and 6c61e6c). For now, put those former three commits onto a separate topic branch, which we can discuss separately.
* Eliminate the `LogLevel` enum changes—unless you still want to discuss further. That change also could be done as a separate PR after this one is completed, though.
* Update `Logger` to include a parent `Logger`. Create a root logger from which all loggers descend. Change log events to propagate up the tree. Eliminate the later series of patches from this PR in favor of root logger subscriptions.
* Polish all commits for style and message format.

If you have time, go ahead and try it. Happy to discuss further as needed, if questions arise. And if you don't have time, let me know that as well, and I can also try to work on it.

src/main/java/org/scijava/console/ConsoleService.java
ctrueden: Ah, I see. You want the `StderrLogService` to write to stdout/stderr, but _without_ publishing `OutputEvent`s. I have mixed feelings about that.

The thing is: in general, listeners of `OutputEvent` want to know about everything that goes to the console—even stuff that originated from the `StderrLogService`. Just because _your_ use case (the fancy Log window) already knows about those messages from the logging side does not mean we should suppress the `OutputEvent` publication globally.

I assume your goal with the fancy log window is to avoid double printing the same message on both the default log channel and then consequently to the console, right? To achieve that, perhaps we should find a way to detect when an `OutputEvent` originated from a logging action, no? Or, the fancy log window could simply tell the truth about all this: if you have a separate tab for the console, it will also show the messages sent to the default logger, and that is OK! No harm done!

maarzt:
> * Eliminate the LogLevel enum changes — unless you still want to discuss further. That change also could be done as a separate PR after this one is completed, if you wish.

There are a lot of reasons why I would like LogLevel to be an enum:
* When working on a logging panel gui component, I realized, designing a gui for filtering and setting of log levels would be much easier, if there is a finit list of valid log levels.
* The API is much easier to understand if a method expects a LogLevel enum, instead of some mystirious interger value.
* Magic numbers like -1, where already used to express that no value was set, but "null" represents this situation much better.
Commit https://github.com/scijava/scijava-common/pull/255/commits/b02194dd1d580393c86213e6a9bd70162b845c95 introduces the LogLevel class, why shouldn't we make it a enum in the first place.

> Cool, it seems like we are converging on a solution, and that this branch is relatively close to that solution. Here are the steps I see still needing to be done:

Great!

> * Change log events to propagate up the tree. Note that this is slightly tricky—think about what happens when you have two child loggers of the same channel but with different users, and one of them performs a logging action—which loggers in the tree get told what, exactly? It may be the case that the child loggers with attached user are special "leaf" loggers, which behave slightly differently than child "subchannel" loggers do.

I don't fully understand the what you intended to describe, but I thought of similar problems. If a plugin acts like this:

```
class TurtleFan {

	@Parameter
	Logger log;

	private LogListener listenerToMyOneLogMessages = ...;

	TurtleFan(Logger log) {
		this.log = log;
		log.addLogListener(listenerToMyOneLogMessages);
	}

	public void run() {
		log.info("I like turtles");
		...
	}

	public static void main(String... args) {
		Logger log = new DefaultLogger();
		TurtleFan alice = new TurtleFan(log);
		alice.run();
		TurtleFan bob = new TurtleFan(log);
		bob.run();
	}
}
```

Here alice.listenerToMyOneLogMessages gets the log messages of bob, although alice wanted to only listen to here one log messages. This situation would be suprising to me. It's caused because Logger provides a bidirectional channel for exchanging messages.
Even worse: There is a memory leak, alice will be kept in memory, only to receive all the log messages from bob. 

That's why I would like to remove addListener, notifyListener, removeLister from the Logger interface, and the DefaultLogger. Instead there should be a ListenableLogger that wraps around a logger and provides this addListener, notifyListener, ... behavior.

The example above would have to be slightly changed:
```
	TurtleFan(Logger log) {
		this.log = new ListenableLogger(log);
		log.addLogListener(listenerToMyOneLogMessages);
	}
```

Currently this memory leak and surprising behavior is quite likely because LogService.channel(...) makes sharing one Logger instance between several clients, easy. The LogService additionally keeps the Loggers and there Listeners in memory.

> If you have time, go ahead and try it. Happy to discuss further as needed, if questions arise. And if you don't have time, let me know that as well, and I can also try to work on it.
I have time to work on this, and I would like to do so.

ctrueden:
> Here alice.listenerToMyOneLogMessages gets the log messages of bob, although alice wanted to only listen to here one log messages. This situation would be suprising to me. It's caused because Logger provides a bidirectional channel for exchanging messages.

Why did you make a constructor for passing the `Logger`? In SciJava, plugins should never have constructors. They have only the implicit no-args constructor. Parameters are injected after construction.

Or is `TurtleFan` not a plugin here? If not, then it should not annotate `Logger` with `@Parameter`. Non-plugin classes of SciJava may only use `@Parameter`s with `Service` and `Context` objects, because those are the only types handled by the core `Context#inject(Object)` injection mechanism.

Also, why did you write `new DefaultLogger()`? Logger instances should always be retrieved from the `LogService` of a context, and never constructed directly.

> Even worse: There is a memory leak, alice will be kept in memory, only to receive all the log messages from bob.

We could solve this by making each `Logger`'s listeners list into weak references instead of strong ones. This is already how the `EventService` works. Maybe it was a mistake to avoid using the `EventService` here for these events; it solves some of the same problems we are now discussing.

> That's why I would like to remove addListener, notifyListener, removeLister from the Logger interface, and the DefaultLogger. Instead there should be a ListenableLogger that wraps around a logger and provides this addListener, notifyListener, ... behavior.

This is basically what I meant by "leaf" logger above. Naming-wise, perhaps we should call the hierarchical thing `LogChannel` instead of `Logger`, and then make `Logger` its own thing which has a `LogChannel` and a `Class<?> userClass` (or `Object user` perhaps) and supports listeners. Calls to `LogService#channel` would return `Logger` instances as discussed earlier, creating a new `LogChannel` if it does not already exist, or wrapping the existing one if it does.

I wanted to write a revised version of your `TurtleFan` example above, but I am honestly not sure what it is trying to achieve, so did not know how best to proceed with it.

maarzt:
I think the Logger interface should be there for passing log messages from a plug in to it's caller.
This addLogListener, notifyLogListener functionality is not needed therefor.

Having the addLogListener and notifyLogListener methods makes using the interface more complicated. In the TurtleFan example I intended to illustrate, that a plug in developer needs to be aware, if the Logger instance is shared between multiple plug ins or not. This means, the plug in needs to know how exactly it is called. This is a dependency between caller and plug in.
Dependency in programming is bad. That's why I would like to remove this addLogListener and notifyLogListener from the Logger interface. I would add a ListanableLogger interface that extends Logger and provides this methods.

You are right, the TurtleFan example is a bad mixture of scijava's `@Parameter` annotation, with non scijava way of passing parameters to the constructor and the non scijava way of constructing a DefaultLogger. I'm sorry for the misleading example. I tried to make it short and easy to understand, but I failed with this.

>> Even worse: There is a memory leak, alice will be kept in memory, only to receive all the log messages from bob.

> We could solve this by making each Logger's listeners list into weak references instead of strong ones. This is already how the EventService works. Maybe it was a mistake to avoid using the EventService here for these events; it solves some of the same problems we are now discussing.

I will have a look at the EventService.

maarzt:
I made a performance test measuring the time a log message needs to be processed with and without retrieving the calling class. See https://github.com/maarzt/scijava-common/commit/f275220ab2ad3bb82d4e48bda6f78bbeca5529ae
This is the result:
```
Time elapsed per log (without calling class): 6.225µs
Time elapsed per log (with calling class): 34.738µs
```
Retrieving the stack trace / calling class took 20µs or more, depending on how much entries there are on the stack.

I did a second performance test to see, which impact the usage of a class LogMessage, would have on the performance. The LogMessage class holds all informations belonging to on log message: level, text, throwable and source, and is used for reducing the number of parameters of methods like Logger::alwaysLog. There is no measurable impact on the performance. See:  https://github.com/maarzt/scijava-common/commit/af3e2260e0b4407240189c04b8adae041e03fea6

I deduce from the performance tests:
1. The decision if a log message is delivered or discarded should not require retrieving the calling class, as this would take to much time especially for "trace" messages, that will most likely be discarded.
2. A LogMessage class should be introduced as it largely increases maintainability and modularity without any performance drawbacks.

@ctrueden How do you feel about making LogLevel a enum in this pull request? It would increase readability of the code and simplify the usage of the API. I think it will be much harder to make this change after this pull request?

ctrueden:
To make it easier for us to work on this together, I have taken the latest work here, rebased it to do some cleanups, and then pushed it as `log-revamp` to this repository. @maarzt I invited you to join the scijava-common team so that you have write access, so we can work on the same branch.

Hence: closing in favor of #272.

PR #256, opened 2017-02-08
Modify cancast

lnyng:
canCast now supports auto-unboxing, and is renamed to canAssigned.

Auto-unboxing is important for conversion using constructors. For example, conversion from `Integer` to `IntType` was not working because the constructor of `IntType` only accepts `int` as parameter.

The renaming is because that this method does not allow "casting" between different primitive types, which is effectively assignment instead.


ctrueden:
:+1: :+1:

PR #257, opened 2017-02-09
Ui console tweaks

ctrueden:
This branch adds the `--showUI` console argument, which allows you to show the UI before other console arguments have been processed.

Thanks to @stelfrich for the idea; see [this Gitter log](https://gitter.im/imagej/imagej/archives/2017/02/09) for related discussion.

### DIFF
--- a/src/main/java/org/scijava/ui/console/HeadlessArgument.java
+++ b/src/main/java/org/scijava/ui/console/HeadlessArgument.java
@@ -28,11 +28,14 @@
  * POSSIBILITY OF SUCH DAMAGE.
  * #L%
  */
-package org.scijava.console;
+package org.scijava.ui.console;
 
 import java.util.LinkedList;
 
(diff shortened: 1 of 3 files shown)


stelfrich:
Works as expected when used with a launcher built from https://github.com/imagej/imagej-launcher/issues/45.

With a version of the launcher built from the referenced PR, raising the `priority` of `ShowUIArgument` is not required (as pointed out by @ctrueden on [Gitter](https://gitter.im/imagej/imagej?at=589dcf5e872fc8ce621a7eb3)).

ctrueden:
So, I tested this PR by doing `fiji --dry-run --showUI --run hello.groovy`, saving the output, then swapping the erroneously ordered arguments. And then it works. So the code here is working—the issue is simply with the ImageJ Launcher. Hence: we can merge this.

PR #259, opened 2017-03-14
Finalize Location / DataHandle API  and add more Locations

gab1one:
This PR iterates on the Location / Datahandle API, stabilizing it to the point that it can be recommended for use.

It adds support for several new Locations:
-  `ByteBankLocation`: based on the new `ByteBank` interface
- `DummyLocation`: for testing purposes

Various support classes and interfaces that facilitate easy implementation of  the `Location` interface


ctrueden:
Please do not rename any existing classes. We can do such renames as part of SJC3. If you want to go a middle road, you could put the new handles into a dedicated subpackage, and leave the existing ones alone. If you _really_ want, you could deprecate the old ones, removing the `@Plugin` annotation from the deprecated copies, but making them trivial extensions of the moved ones in the `.handle` subpackage. And please use singular tense for the subpackage (`.handle`, not `.handles`)—this is consistent with all the other subpackages in SJC.

ctrueden:
FYI, I am planning to review within the next few days. Goal is to release a new SciJava Common which includes it before end of next week.

src/main/java/org/scijava/io/AbstractDataHandle.java
```diff
@@ -71,4 +77,23 @@ public void setEncoding(final String encoding) {
 		this.encoding = encoding;
 	}
 
+	@Override
+	public void writeUTF(String str) throws IOException {
```
ctrueden: This is not a new default method of `DataHandle`. I.e.: this change does not match the commit message. It should be done separately.

src/main/java/org/scijava/io/BytesHandle.java
```diff
@@ -231,26 +231,31 @@ public short readShort() throws IOException {
...
 	{
-		validateLength(len);
+		ensureWritable(len);
 		bytes().put(b, off, len);
 	}
 
 	@Override
 	public void write(final int b) throws IOException {
-		validateLength(1);
+		writeByte(b);
```
ctrueden: Confusing that these changes to `BytesHandle` happen in a commit purportedly about `DataHandle` default methods.

src/main/java/org/scijava/io/DataHandle.java
```diff
@@ -42,8 +44,8 @@
 import org.scijava.plugin.WrapperPlugin;
 
 /**
- * A <em>data handle</em> is a plugin which provides access to bytes in a data
- * stream (e.g., files or arrays), identified by a {@link Location}.
+ * A <em>data handle</em> is a plugin which provides both streaming and random
+ * access to bytes at a {@link Location} (e.g., files or arrays).
```
ctrueden: :+1: Much better.

src/main/java/org/scijava/io/DataHandle.java
```diff
@@ -74,43 +82,86 @@
...
 	/**
+	 * Verifies that the handle has sufficient bytes available to read, returning
+	 * the actual number of bytes which will be possible to read, which might
+	 * be less than the requested value.
+	 * 
+	 * @param count Number of bytes to read.
+	 * @return The actual number of bytes available to be read.
+	 * @throws IOException If something goes wrong with the check.
+	 */
+	default long available(final long count) throws IOException {
```
ctrueden: I almost wrote here that we shouldn't name this method `available`. The reason being that for `InputStream`, `available` does _not_ equate to "number of bytes till EOF" but rather "number of bytes which could currently be read from the stream." However, I think the intent there is that you can have a stream which is producing bytes in real time—you may read up to the _current_ end of stream, but more might appear later. So actually, the question really is: __Do we want to support increasingly-long `Location` types? Or only statically sized `Location`s?__ The original use case definitely assumes only statically sized things—i.e., constant `length()`. I think it is safest and best to retain this assumption, and document it carefully. Do you agree @gab1one?

ctrueden: I guess for writing, the length can of course change. My comments above apply to reading. Hmm, thoughts?

src/main/java/org/scijava/io/DataHandle.java
```diff
@@ -74,43 +82,86 @@
...
+		final long minLength = offset() + count;
+		if (length() < minLength) {
+			setLength(minLength);
 			return false;
 		}
 		return true;
 	}
 
+	/** Returns the current order of the stream. */
+	ByteOrder getOrder();
```
ctrueden: Not a `default` method. Again, should be a separate commit from the intent stated above.

src/main/java/org/scijava/io/DataHandle.java
```diff
@@ -74,43 +82,86 @@
...
+
 	/**
-	 * Returns the current order of the stream.
+	 * Returns true iff the stream's order is {@link ByteOrder#BIG_ENDIAN}.
 	 * 
-	 * @return See above.
+	 * @see #getOrder()
 	 */
-	ByteOrder getOrder();
+	default boolean isBigEndian() {
```
ctrueden: :+1:

src/main/java/org/scijava/io/DataHandle.java
ctrueden: Oh I see, the diff is just funny due to other inserted code and/or moved code blocks. That can be avoided by doing more granular commits.

src/main/java/org/scijava/io/BytesLocation.java
```diff
@@ -35,8 +35,14 @@
 
 /**
  * {@link Location} backed by a {@link ByteBuffer}.
- *
+ * <p>
+ * Note that the backing {@link ByteBuffer} reference is mutable, and might
```
ctrueden: :+1:

src/main/java/org/scijava/io/Location.java
```diff
@@ -42,7 +42,7 @@
  * resource identifier</a> ({@link URI}), a location identifies <em>where</em>
  * the data resides, without necessarily specifying <em>how</em> to access that
  * data. The {@link DataHandle} interface defines a plugin that knows how to
- * provide a stream of bytes for a particular kind of location.
+ * read and/or write bytes for a particular kind of location.
```
ctrueden: :+1: Better. Also because it's not just about streams, but random access too.

src/main/java/org/scijava/io/DataHandle.java
```diff
@@ -82,12 +82,31 @@
...
+	 * <p>
+	 * In the case of writing, attempting to write the returned number of bytes is
+	 * guaranteed not to expand the length of the handle; i.e., the write will
+	 * only overwrite bytes already within the handle's bounds.
+	 * </p>
 	 * 
-	 * @param count Number of bytes to read.
-	 * @return The actual number of bytes available to be read.
+	 * @param count Desired number of bytes to read/write.
+	 * @return The actual number of bytes which could be safely read/written,
```
ctrueden: I'd remove "safely" here. It's not about safety, it's about feasibility. You cannot read past the end of the data.

src/main/java/org/scijava/io/BytesHandle.java
```diff
@@ -116,6 +116,18 @@ public void seek(final long pos) throws IOException {
...
 
+	@Override
+	public boolean isReadable() {
+		return true;
+	}
+
+	@Override
+	public boolean isWritable() {
+		// NB: Actually, some ByteBuffers are read-only.
+		// But sadly there is no API for determining this.
```
ctrueden: Ah, but we _can_ make a best guess. I suggest:
```
return !(this instanceof HeapByteBufferR || this instanceof DirectByteBufferR)
```
as a heuristic that gets us 99% of the way there. No?

ctrueden:
I did a few passes over the history. It is getting closer:

* The work on `DataHandle` now happens first over many commits.
* Then work on `FileLocation` and `FileHandle`.
* Then work on `BytesLocation` and adding the new `BytesHandle`.
* Then a commit adding read-only and write-only distinction.
* Then adding dummy location and handle.
* Then adding `StreamHandle` and friends.
* Then adding `IOBuffer` and a `ByteArray`-based implementation thereof.
* Finally, the WIPs—work which needs further discussion and consideration, IMHO.

Still to do, consider, discuss:

1. Split each location/handle pair into its own subpackage. Two issues there: A) What to name the packages—e.g., `FileLocation` and `FileHandle` go in `org.scijava.io.location.file`? Or just `org.scijava.io.file`? Or perhaps` org.scijava.io.handle.file`. And B) What to do about the existing `FileLocation` and `FileHandle`? Copy and deprecate the old ones? Or just leave them alone?

2. Improve the `IOBuffer`. It should also support working with `ByteBuffer` in its API and not just `byte[]`. Then it will be useful for copying to/from off-heap data as well. And there should be another implementation backed by a `ByteBuffer` instead of a `ByteArray`.

3. Deprecate `BytesLocation` in favor of `IOBufferLocation`. Change the `BytesHandle` to instead be an `IOBufferHandle`. With this change, we have a level of indirection which prevents the yuckiness of `BytesLocation` having a mutable backing `ByteBuffer`. Instead, it has a stable backing `IOBuffer`, which has a mutable `ByteBuffer`. This is a more acceptable structure. (We can have convenience constructors for making `IOBufferLocation`s from `ByteBuffer` directly, in the same way that `FileLocation` has convenience constructors letting you bypass `java.io.File` construction.)

4. Discuss the streaming and browsing APIs. The browsing API in particular I think should maybe be a separate `Browseable` (or maybe `BrowseableLocation`) interface, rather than baked directly into `Location`. We may want to merge this branch with everything up to the `DummyLocation` and `DummyHandle`, and keep the commits from `RemoteLocation` onwards on another topic branch until its sorted out a bit more.

Bed time for me now, and then it's notebook prep time for the Learnathon tomorrow, so it may be a few days before I return to this again. In the meantime @gab1one, I'm happy to hear what you think, agreement, disagreement, ideas, etc.

src/main/java/org/scijava/io/DataHandle.java
gab1one: true, I changed that.

src/main/java/org/scijava/io/DataHandle.java
gab1one: I think this already supports streams that increase in size, as we do not cache the size. It just ensures you can don't get an exception when reading. But I agree it needs a more descriptive name, maybe something like ``clampToAvailableBytes`` ?

gab1one:
1. In principle this idea prevents us cluttering the `handles` package if we add
   more handles in the future, so we should implement. I think
   `org.scijava.io.handles.file` is my favorite for the package naming. 
   If we do not move the old handles, we end up with the random deprecated classes
   scattered over the hierarchy, which I would like to avoid. I doubt
   the handles are used anywhere currently, or am I wrong with that assumption?
   If they are unused downstream we can just make the breaking change of moving
   them.

2. That should be quite straightforward to do.

3. That is an interesting idea, maybe we can also just remove the
   `setByteBuffer()` method, because it violates the concept of `Locations` as
   more immutable descriptors. Anyway, having an `IOBufferHandle` is useful in any case. 
   
4. I baked the browsing methods directly into the `Location` to avoid casting 
   when I consumed them in SCIFIO. When we move them into a dedicated interface
   we need to do a lot of `instance of` checks. I guess this is more of an taste
   question. I think independently of the other browsing methods, `exists()` should be
   part of the Location API.

ctrueden:
I rewrote the branch again. It is still not ready. I will keep massaging it now. Hopefully we can get (most of) this merged by tomorrow.

ctrueden:
@gab1one I am done rebasing! I am happy with the current status now.

What I would like to do is move the last three commits—the ones adding `StreamHandle` and subtypes—to a separate topic branch, and then merge everything else here to master. Could you take a look and tell me if you have any concerns?

Note that I decided to put all `Location` types into `org.scijava.io.location`, and all `DataHandle` types into `org.scijava.io.handle`, because it is fewer packages, and less nested. The base `org.scijava.io` stays clean—especially after we remove the defunct `RecentFileService`. I hope this decision does not bother you too much.

ctrueden:
@gab1one OK for you if I merge everything here except for the final 3 commits?

ctrueden:
I take it back; there was another wave of changes I wanted to make, which I have written in my notes, but which I forgot to make. Doing it now.

ctrueden:
Made the wave of changes. I now have the following two questions / TODOs for you, @gab1one:

1. In 703b48d07373f319b7a60736b0dffa253b46f8cc ("DataHandle: tweak readString, skip and skipBytes"), the `skipBytes` method of `DataHandle` is still _not_ leaning on `available(long)` as advertised. Should it?

2. In 911517ea44f4d62b480c63f8ea57dd12506e57f0 ("FileHandle: implement getName()"), should we use `getAbsolutePath()` rather than `getName()`?

gab1one:
1. It definitely should that is a good catch!

2. The `getName()` method is actually part of the file browsing changes, it is intended to provide the name of the adressed file in it's local context, not the full path. The usecase is to provide a human readable name  that is independend of the storage location. I used it in Scifio in a few places, because I did not want to rely on extracting the basepath / filename from the URL. Maybe this method needs a better and a different default implementation.

gab1one:
Looking at the available Locations / Handles, should we really include `URILocation` and `URLLocation` if there are no Handles that are based on them? There is not really a reason to implement them if we do not know yet if they are matching the handles.

ctrueden:
> It definitely should that is a good catch!

Updated the history to fix this.

> Maybe this method needs a better and a different default implementation.

How about `getShortName()` and `getFullName()`? Or is that too convoluted?

gab1one:
For HTTP URLs there is the following external project, https://github.com/gab1one/scijava-handles-http
We decided to not include it into scijava-common because of it's extra dependencies. It has it's own `HTTPLocation` which supports different URL formats including `java.net.URL `

gab1one:
How about `getShortName()` and `getFullName()`? 
- `getShortName()` could be `getFileName()`? To express it is based on path information locations (e.g. ByteBank) 
- `getFullName()` is equivalent to `getURI()` in my eyes.

ctrueden:
> `getFullName()` is equivalent to `getURI()` in my eyes.

Good point. We can stick with `getName()` for the short name. I am going to add some javadoc to that effect.

ctrueden:
Merged. See `stream-locations` branch for the `RemoteLocation` and various `StreamLocation` commits on top of this merged work.

PR #264, opened 2017-04-18
Add autocompletion framework

hadim:
Add a framework for an autocompletion framework. Each scripting language is free to implement an `AutoCompleter` to provide autocompletion.

If they don't the default `AdaptedAutoCompleter` will just return no results when an autocompletion is asked.

### DIFF
--- a/src/main/java/org/scijava/script/AbstractScriptLanguage.java
+++ b/src/main/java/org/scijava/script/AbstractScriptLanguage.java
@@ -31,6 +31,7 @@
 
 package org.scijava.script;
 
+import javax.script.ScriptEngine;
 import javax.script.ScriptEngineFactory;
 
 import org.scijava.plugin.AbstractRichPlugin;
@@ -72,11 +73,6 @@ public abstract class AbstractScriptLanguage extends AbstractRichPlugin
 		if (info != null) name = info.getName();
 		return name != null && !name.isEmpty() ? name : inferNameFromClassName();
 	}
-    
-    @Override
-    public AutoCompleter getAutoCompleter() {
-        return new DefaultAutoCompleter(getScriptEngine());
-    }
 
 	// -- Helper methods --
 
--- a/src/main/java/org/scijava/script/autocompletion/AbstractAutoCompleter.java
+++ b/src/main/java/org/scijava/script/autocompletion/AbstractAutoCompleter.java
@@ -30,15 +30,13 @@
  */
 package org.scijava.script.autocompletion;
 
-import java.lang.reflect.Array;
 import java.lang.reflect.Field;
-import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
-import java.util.Map;
+import java.util.stream.Collectors;
 import javax.script.Bindings;
 import javax.script.ScriptContext;
 import javax.script.ScriptEngine;
@@ -82,7 +80,12 @@ public abstract class AbstractAutoCompleter implements AutoCompleter {
             matches.addAll(this.engineVariablesCompleter(code, index, engine));
         }
 
-        // Sort matches alphabetcially
+        // Remove duplicates
+        matches.stream()
+                .distinct()
+                .collect(Collectors.toList());
+
+        // Sort alphabetcially
         Collections.sort(matches, new SortIgnoreCase());
 
         // Return results. For now we ignore index and startIndex.
--- a/src/main/java/org/scijava/script/autocompletion/AbstractAutoCompleter.java
+++ b/src/main/java/org/scijava/script/autocompletion/AbstractAutoCompleter.java
@@ -28,67 +28,40 @@
  * POSSIBILITY OF SUCH DAMAGE.
  * #L%
  */
+package org.scijava.script.autocompletion;
 
-package org.scijava.script;
-
-import javax.script.ScriptEngineFactory;
-
-import org.scijava.plugin.AbstractRichPlugin;
-import org.scijava.plugin.PluginInfo;
-import org.scijava.script.autocompletion.AutoCompleter;
-import org.scijava.script.autocompletion.DefaultAutoCompleter;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+import javax.script.ScriptEngine;
 
 /**
- * Abstract superclass for {@link ScriptLanguage} implementations.
- * <p>
- * This class implements dummy versions of {@link ScriptEngineFactory}'s methods
- * that are not needed by the SciJava scripting framework.
- * </p>
- * 
- * @author Johannes Schindelin
+ *
+ * @author Hadrien Mary
  */
-public abstract class AbstractScriptLanguage extends AbstractRichPlugin
-	implements ScriptLanguage
-{
+public abstract class AbstractAutoCompleter implements AutoCompleter {
 
-	// -- Object methods --
+    protected ScriptEngine engine = null;
 
-	@Override
-	public String toString() {
-		return getLanguageName();
-	}
-
-	// -- Default implementations --
-
-	@Override
-	public String getEngineName() {
-		return inferNameFromClassName();
-	}
-
-	@Override
-	public String getLanguageName() {
-		String name = null;
-		final PluginInfo<?> info = getInfo();
-		if (info != null) name = info.getName();
-		return name != null && !name.isEmpty() ? name : inferNameFromClassName();
-	}
-    
-    @Override
-    public AutoCompleter getAutoCompleter() {
-        return new DefaultAutoCompleter(getScriptEngine());
+    public AbstractAutoCompleter(ScriptEngine engine) {
+        this.engine = engine;
     }
 
-	// -- Helper methods --
+    @Override
+    public Map<String, Object> autocomplete(String code) {
+        return autocomplete(code, 0);
+    }
+    
+    @Override
+    public Map<String, Object> autocomplete(String code, int i) {
+        Map<String, Object> result = new HashMap<>();
+        result.put("matches", new ArrayList<>());
+        result.put("startIndex", 0);
+        return result;
+    }
 
-	private String inferNameFromClassName() {
-		String className = getClass().getSimpleName();
-		if (className.endsWith("ScriptLanguage")) {
-			// strip off "ScriptLanguage" suffix
-			className = className.substring(0, className.length() - 14);
-		}
-		// replace underscores with spaces
-		className = className.replace('_', ' ');
-		return className;
-	}
-
+    @Override
+    public ScriptEngine getScriptEngine() {
+        return this.engine;
+    }
 }
--- a/src/main/java/org/scijava/script/autocompletion/AbstractAutoCompleter.java
+++ b/src/main/java/org/scijava/script/autocompletion/AbstractAutoCompleter.java
@@ -30,10 +30,15 @@
  */
 package org.scijava.script.autocompletion;
 
+import java.lang.reflect.Array;
+import java.lang.reflect.Field;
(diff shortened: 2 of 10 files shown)


hadim:
@ctrueden let me know what do you think. Even if autocompletion is not implemented yet in the scripting languages, it's fine since all the changes are backward compatible.

ctrueden:
Great, this is something I've been wanting to add to the SciJava script framework. (The other major missing thing is a "script recorder" type feature, which given a SciJava module execution, generates the line of code needed to reproduce it in that script language.)

I will try to find time to review the details later this week. Ping me again if I'm holding you up.

hadim:
I don't think I'll have time to implement autocompletion for scripting languages any time soon (need at least another free weekend :-)) but as I said this PR should hold the main foundation of autocompletion.

So feel free to merge after reviewing it.

I'll ping you here at the end of this weekend :-)

hadim:
Also once it's done that work could be used in the Scijava TextEditor :-D

imagejan:
Just for reference: [Icy](http://imagej.net/Icy) has a nice auto-completion here:
https://github.com/tprovoost/Script-Editor/tree/master/javascripter/src/plugins/tprovoost/scripteditor/completion

It was discussed shortly at NEUBIAS2017 and suggested by @ctrueden that this could be ported into https://github.com/scijava/script-editor

It didn't look deep into this code, but maybe we could still profit from their implementations for JS and Python...

hadim:
Well, I am surprised to see how easy it was to implement a first "naive" working autocompletion !

That implementation should provide autocompletion for all the scripting languages. 

I am pretty happy with that for now. The framework should provide an easy way to extend autocompletion being language specific or not.

About the architecture, I am sure it can be improved (I did my best but IJ coding standard are pretty high ! (which is a good thing)). So I am waiting for @ctrueden comments about that :-)

src/test/java/org/scijava/script/ScriptInfoTest.java
```diff
@@ -285,6 +287,11 @@ public ScriptEngine getScriptEngine() {
 		public List<String> getExtensions() {
 			return Arrays.asList("bsizes");
 		}
+
+        @Override
+        public AutoCompleter getAutoCompleter(ScriptEngine scriptEngine) {
+            return new DefaultAutoCompleter(this, scriptEngine);
```
ctrueden: This override is unnecessary, no? Because `AbstractScriptLanguage` already defines this same logic?

src/test/java/org/scijava/script/ScriptFinderTest.java
```diff
@@ -241,6 +243,11 @@ public ScriptEngine getScriptEngine() {
 			throw new IllegalStateException();
 		}
 
+        @Override
+        public AutoCompleter getAutoCompleter(ScriptEngine scriptEngine) {
+            return new DefaultAutoCompleter(this, scriptEngine);
```
ctrueden: This override is unnecessary, no? Because `AbstractScriptLanguage` already defines this same logic?

src/test/java/org/scijava/script/ScriptEngineTest.java
```diff
@@ -99,6 +101,11 @@ public ScriptEngine getScriptEngine() {
 		public List<String> getExtensions() {
 			return Arrays.asList("rot13");
 		}
+
+        @Override
+        public AutoCompleter getAutoCompleter(ScriptEngine scriptEngine) {
+            return new DefaultAutoCompleter(this, scriptEngine);
```
ctrueden: This override is unnecessary, no? Because `AbstractScriptLanguage` already defines this same logic?

src/test/java/org/scijava/menu/MenuServiceTest.java
```diff
@@ -68,6 +68,6 @@ public void tearDown() {
 	 */
 	@Test
 	public void testMenuCreation() {
-		menuService.getMenu();
+		//menuService.getMenu();
```
ctrueden: ???

src/main/java/org/scijava/script/autocompletion/AbstractAutoCompleter.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Override
+    public ScriptEngine getScriptEngine() {
+        return this.scriptEngine;
+    }
+
+    @Override
+    public ScriptLanguage getScriptLanguage() {
+        return this.scriptLanguage;
```
ctrueden: Nothing actually uses this field in the base layer... do you use it in downstream implementations?

src/main/java/org/scijava/script/autocompletion/AbstractAutoCompleter.java
```diff
@@ -0,0 +1,111 @@
...
+        // Sort matches alphabetcially
+        Collections.sort(matches, new SortIgnoreCase());
+
+        // Return results. For now we ignore index and startIndex.
+        return new AutoCompletionResult(matches, startIndex);
+    }
+
+    @Override
+    public ScriptEngine getScriptEngine() {
+        return this.scriptEngine;
```
ctrueden: Alternately, the autocompleter could be kept stateless, with the `ScriptEngine` instance being passed as another argument to the `autocomplete` methods. Which do you like better? (I don't super strongly; just pointing it out.)

src/main/java/org/scijava/script/AbstractScriptLanguage.java
```diff
@@ -70,6 +73,11 @@ public String getLanguageName() {
 		if (info != null) name = info.getName();
 		return name != null && !name.isEmpty() ? name : inferNameFromClassName();
 	}
+    
+    @Override
+    public AutoCompleter getAutoCompleter(ScriptEngine scriptEngine) {
+        return new DefaultAutoCompleter(this, scriptEngine);
```
ctrueden: Now that we have Java 8, this could be a `default` method of `ScriptLanguage` interface itself, rather than in the abstract base class. That would potentially be "more backwards compatible", just in case anyone out there has implemented the `ScriptLanguage` interface directly.

src/test/java/org/scijava/menu/MenuServiceTest.java
ctrueden: This needs more explanation.

ctrueden:
I made some comments. The only real concern I have is that failing test... otherwise, all looks great. We should decide once and for all if `Autocompleter` will be stateless, or will wrap a `ScriptEngine` instance the way you have now. Once we decide that, I'll merge it, and do some minor subsequent cleanup.

ctrueden:
Relating to @imagejan's comments about Icy's autocompletion: one thing that Icy does is that it gives special priority to the Icy-specific classes. So you end up getting completions that are relevant to the context of Icy plugin development.

The current framework is very general, which is good, but I'm wondering what the best way forward is to make it possible to give contextual hints like that, but in an extensible way. (Icy just hardcodes everything.) One option would be to have some kind of new plugin type that identifies high-priority completions. It would be nice if there were not language-specific either, but rather a layer which operates sort of orthogonally with the actual completion strings. Maybe something as simple as "with such-and-such prefix, which classes make sense to autocomplete?" That will be the same across all script languages, and it will allow components to highlight "completion-friendly" classes. All of these thoughts are half-baked, of course, but wanted to jot this down before it's out of my head again.

src/test/java/org/scijava/script/ScriptInfoTest.java
hadim: Netbeans does not agree with you (only a warning). Should I just remove it ?

src/test/java/org/scijava/menu/MenuServiceTest.java
hadim: I don't know why this test was failing. This is not the case anymore. I'll fix that.

src/main/java/org/scijava/script/autocompletion/AbstractAutoCompleter.java
hadim: Nope. I'll remove it.

src/main/java/org/scijava/script/AbstractScriptLanguage.java
hadim: I'll do that.

src/test/java/org/scijava/script/ScriptInfoTest.java
ctrueden: Weird. What is the warning? The whole point of having methods implemented in the abstract base class is so that you _don't_ have to implement them that way same in the subclass. NetBeans warning about that does not make sense...

src/main/java/org/scijava/script/autocompletion/AbstractAutoCompleter.java
hadim: Good point here and I like the idea. That will make the class less complicated for nothing.

src/test/java/org/scijava/script/ScriptInfoTest.java
hadim: It's saying `Add @Override Annotation`. Note that the other methods above has `@Override`

imagejan: IIUC, @ctrueden means you can leave away the whole method, not just the annotation, no?

ctrueden: Yes, just delete the whole overridden method. Lines 291 through 294.

hadim: Oh ok !

hadim:
All the comments have been addressed. Feel free to merge this one and we can still extend it later. I think it's good enough to start with.

About Icy code, I had a look but I didn't find useful things to steal so I just started from scratch.

The framework is simple for now but I definitely want to make something extensible. For example, we could have `AutoCompleterPlugin` discovered at runtime and used by `AutoCompleter`.

My last commit implements autocompletion by inspecting attributes/methods of an object (detected if the string ends by `.`). That work pretty well but it's not perfect. And of course reflection does not work with Jython (didn't try very hard).

hadim:
Thank you both for your comments.

hadim:
Once this has been merged, I'll cut a new release of the kernel, build conda packages and make an announcement on the forum (I'll also need a new `scijava-common` release).

ctrueden:
Do you need a new `pom-scijava`, or is a new `scijava-common` sufficient?

I am currently reworking how [status.imagej.net](https://status.imagej.net) works (actually, it will become `status.scijava.org`), since the switch to Travis CI broke the old way of generating that page.

hadim:
Well, I just need the last `scijava-common` version (to get this PR). But I guess you need to edit the version in the pom ?

ctrueden:
Eventually we need to update the pom-scijava bill of materials to reference the new SciJava Common version. But in the meantime, you can likely override the `scijava-common.version` property to point to the forthcoming release.

I just asked because I can merge and release a new SJC soon (will try this weekend if I have time), but getting all my ducks in a row to do the full suite of releases and update pom-scijava to match is a larger endeavor which will take a few more days at least.

hadim:
Oh, I see. So if you just cut a new release of `scijava-common`, it's fine. I'll override the version property.

ctrueden:
Just a note that I haven't forgotten about this. I am planning to clean up and merge this branch soon (within a few days). I have to sort out how to make releases to OSS Sonatype though, now that our Jenkins configuration is hosed up. I want to instead do it from the CLI, without needing Jenkins anymore. Once I have that sorted, I'll try to get this merged, and then cut a new `scijava-common` release.

hadim:
Thanks @ctrueden .

If you also have time to look at #265 , that would be nice ! Both PRs carry the final features before having a pretty stable scijava kernel.

Releasing from CLI is definitely the way to go. Good luck with that.

ctrueden:
Cleaned up, fleshed out and merged!

PR #266, opened 2017-05-15
Clarify missing input message

rimadoma:
Make the message about missing inputs clearer in case of images. An ImagePlus won't be
considered an image, but referencing the class would create a circular dependency.


ctrueden:
Sorry, but `scijava-common` cannot depend on ImgLib2. The entire point of the SciJava layer is to be image-agnostic. We would need to come up with some other way of improving the error message.

rimadoma:
Fair enough 👍

PR #271, opened 2017-06-16
Handle Plugins*Event in AbstractSingletonService

stelfrich:
For most `SingletonServices` the `instances` and `instanceMap` fields were only set during initialization making them inflexible during runtime. Registering to `PluginAddedEvents` and `PluginRemovedEvents` allows for more dynamic behavior, e.g. adding `Converter`s in notebooks.

/cc @ctrueden

### DIFF
--- a/src/main/java/org/scijava/plugin/AbstractSingletonService.java
+++ b/src/main/java/org/scijava/plugin/AbstractSingletonService.java
@@ -33,11 +33,16 @@ package org.scijava.plugin;
 
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 
+import org.scijava.event.EventHandler;
+import org.scijava.event.EventService;
 import org.scijava.log.LogService;
 import org.scijava.object.ObjectService;
+import org.scijava.plugin.event.PluginsAddedEvent;
+import org.scijava.plugin.event.PluginsRemovedEvent;
 
 /**
  * Abstract base class for {@link SingletonService}s.
@@ -55,8 +60,8 @@ public abstract class AbstractSingletonService<PT extends SingletonPlugin>
 	@Parameter
 	private ObjectService objectService;
 
-	// TODO: Listen for PluginsAddedEvent and PluginsRemovedEvent
-	// and update the list of singletons accordingly.
+	@Parameter
+	private EventService eventService;
 
 	/** List of singleton plugin instances. */
 	private List<PT> instances;
@@ -73,7 +78,7 @@ public abstract class AbstractSingletonService<PT extends SingletonPlugin>
 	@Override
 	public List<PT> getInstances() {
 		if (instances == null) initInstances();
-		return instances;
+		return Collections.unmodifiableList(instances);
 	}
 
 	@SuppressWarnings("unchecked")
@@ -88,8 +93,9 @@ public abstract class AbstractSingletonService<PT extends SingletonPlugin>
 	private synchronized void initInstances() {
 		if (instances != null) return;
 
-		final List<PT> list = Collections.unmodifiableList(filterInstances(
-			pluginService().createInstancesOfType(getPluginType())));
+		@SuppressWarnings("unchecked")
+		final List<PT> list = (List<PT>) filterInstances(pluginService()
+			.createInstancesOfType(getPluginType()));
 
 		final HashMap<Class<? extends PT>, PT> map =
 			new HashMap<>();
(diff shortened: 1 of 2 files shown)


src/main/java/org/scijava/plugin/AbstractSingletonService.java
```diff
@@ -108,4 +114,48 @@ private synchronized void initInstances() {
...
+				list.add(instance);
+			}
+			catch (ClassCastException exc) { /* NB */ }
+		});
+
+		filterInstances(list).forEach(instance -> {
+			@SuppressWarnings("unchecked")
+			final Class<? extends PT> ptClass = (Class<? extends PT>) instance
+				.getClass();
+			instanceMap.putIfAbsent(ptClass, instance);
```
ctrueden: If someone tries to add a plugin which already exists, we don't overwrite it?

src/test/java/org/scijava/plugin/SingletonServiceTest.java
```diff
@@ -0,0 +1,131 @@
...
+	public void tearDown() {
+		pluginService.context().dispose();
+	}
+
+	/**
+	 * Tests that the {@link AbstractSingletonService} properly handles
+	 * {@link PluginsAddedEvent}s originating from the {@link PluginService}.
+	 */
+	@Test
+	public void testSingletonServicePluginsAddedHandling() {
```
ctrueden: :+1:

src/test/java/org/scijava/plugin/SingletonServiceTest.java
```diff
@@ -0,0 +1,131 @@
...
+		@SuppressWarnings("rawtypes")
+		PluginInfo<Converter> converterInfo = new PluginInfo<>(
+			DummyStringConverter.class, Converter.class);
+		converterInfo.setPluginClass(DummyStringConverter.class);
```
ctrueden: Is it actually necessary to call `setPluginClass` here? The constructor should do it?

src/test/java/org/scijava/plugin/SingletonServiceTest.java
```diff
@@ -0,0 +1,131 @@
...
+	/**
+	 * Tests that the {@link AbstractSingletonService} properly handles
+	 * {@link PluginsRemovedEvent}s originating from the {@link PluginService}.
+	 */
+	@Test
+	public void testSingletonServicePluginsRemovedHandling() {
+		@SuppressWarnings("rawtypes")
+		PluginInfo<Converter> converterInfo = new PluginInfo<>(
+			DummyStringConverter.class, Converter.class);
+		converterInfo.setPluginClass(DummyStringConverter.class);
```
ctrueden: Same here: is this call necessary?

src/test/java/org/scijava/plugin/SingletonServiceTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+		pluginService.addPlugin(converterInfo);
+
+		// De-register DummyStringConverter
+		pluginService.removePlugin(converterInfo);
```
ctrueden: Instead of adding then immediately removing, a cleaner approach might be to simply remove one of the built-in discovered plugins. If you don't like that external dependency for some reason (and I wouldn't blame you), you could create a second static inner class here annotated with `@Plugin` so that it is automatically discovered and added initially, and then remove it here. The advantage is that we ensure it works to remove plugins where were present _when the context was first initialized_. I think this is a more robust test—or at least a needed additional test.

src/test/java/org/scijava/plugin/SingletonServiceTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+		assertNull(pluginService.getPlugin(DummyStringConverter.class));
+		assertFalse(convertService.supports("StringInput", Number.class));
```
ctrueden: Wait... doesn't the convert service support conversion of string to number? Or only if you give an explicit `Number` subtype otherwise?

I would prefer, instead of `String` and `Number`, that you use two new interfaces defined here. That way we can be _certain_ that future converters relating to basic types like strings and numbers do not inadvertently cause this test to fail in the future.

ctrueden:
Thank you very much, @stelfrich, this is looking really good. I'm glad it turned out to be straightforward and clean to implement.

I made some comments—could you please address them if you have time? Should be relatively minor and quick I hope.

stelfrich:
> I made some comments—could you please address them if you have time? Should be relatively minor and quick I hope.

I have addressed your comments @ctrueden.

ctrueden:
Meanwhile, @gab1one did this work again in ecf6cbb381b7fbaa6fb007aea388dc48ef2baaed. :sweat: @gab1one: will you please take a look at this PR and see if it accomplished anything that your changes did not?

@stelfrich I am very sorry that I not only dropped the ball on merging this, but then forgot it even existed. I am now making it a priority to go through the substantial backlog of very worthy PRs across the various SciJava and ImageJ components.

stelfrich:
> I am very sorry that I not only dropped the ball on merging this, but then forgot it even existed.

Don't worry! I know how hard it is to keep track of all the PRs...

>  @gab1one: will you please take a look at this PR and see if it accomplished anything that your changes did not?

Just go ahead and close this if there's nothing of additional value here (which is likely)..

PR #274, opened 2017-07-07
Port support for Compressed Handles from scifio

gab1one:
This PR adds support for compressed handles, that operate on archives: namely `.zip`, `.gzip` and `bzip2` archives.

This is based on https://github.com/scijava/scijava-common/pull/259 which needs to be merged first.


gab1one:
Look at Apache commons compress to improve this

ctrueden:
* We should consider using [Apache Commons Compress](https://commons.apache.org/proper/commons-compress/) to read (and maybe write, though less important) these formats and more.
* The locations should implement `BrowsableLocation` (as currently on master) such that code that works on arbitrary `BrowsableLocation` can work with no changes with these compressed formats.
* `ZipEntry` is a lot like `File`, including stuff like last modified datestamp, which corresponds to `DataHandle` in our paradigm. So a `ZipEntryHandle` would make sense.
* The `ZipLocation` points at a `Location` corresponding to the ZIP data. Its `parent()` method returns `null`. (It _could_ return the containing folder of the zip, but this has pros and cons. We should perhaps experiment with that.)
* A `ZipElementLocation` (naively, `ZipEntryLocation`, but it does _not_ use `java.util.zip.ZipEntry`!) would have a parent `ZipLocation` together with a `String` for the entry name. It is not stateful.

ctrueden:
@gab1one What is the status of this PR now?

ctrueden:
I rebased this over the latest master, removing one commit that was obsolete (which fixed the handling of unknown lengths in an incompatible way with a different fix that was already merged). @hinerm I'd value your input on this PR at some point when you have time. No rush though.

PR #276, opened 2017-07-09
Add method signatures and preprocessor to select multiple files

imagejan:
This PR adds the `chooseFiles(File[] files, FileFilter filter)` and `chooseFiles(List<File> fileList, FileFilter filter)` method signatures.

It is the base for implementations of a file list choosing widget implemented in `scijava-ui-swing`.


imagejan:
I just rebased this on top of current `master`. Builds fine with passing tests.

The issue of persisting `File[]` values stored as `String`s will be addressed in a follow-up PR.

ctrueden:
Thanks!

PR #277, opened 2017-07-13
Add RemoteLocation interface and abstract class

gab1one:
A RemoteLocation is not located on the current machine. We need the
ability to set a connection timeout for such locations, to prevent them
from halting the whole program.


imagejan:
Instead of the abstract class, could you have defined `default` implementations directly in the interface? Does that make any difference functionally?

PR #280, opened 2017-07-18
Bytebank tests and bugfixes

gab1one:
Adds tests for ByteBank and fixes bugs uncovered with these tests

### DIFF
--- a/src/test/java/org/scijava/io/nio/ByteBufferByteBankTest.java
+++ b/src/test/java/org/scijava/io/nio/ByteBufferByteBankTest.java
@@ -35,15 +35,41 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
 import java.nio.ByteBuffer;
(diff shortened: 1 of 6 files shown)


src/test/java/org/scijava/io/nio/ByteBufferByteBankTest.java
```diff
@@ -35,15 +35,41 @@
...
+	@Parameter
+	public Function<Integer, ByteBuffer> supplier;
+
+	@Parameters
+	public static Object[] params() {
+		final Function<Integer, ByteBuffer> alloc = ByteBuffer::allocate;
+		final Function<Integer, ByteBuffer> allocDirect =
+			ByteBuffer::allocateDirect;
+		return new Function[] { alloc, allocDirect };
+	}
```
gab1one: This allows us to test both ByteBuffer variants in one unit test.

ctrueden:
Thanks!

PR #284, opened 2017-07-20
Add Tests to Ensure Proper Converter Chosen

awalter17:
Hello,

This branch makes the following changes:
* Switches `DefaultConverter` to `VERY_LOW_PRIORITY - 1000`, `NullConverter` to `VERY_HIGH_PRIORITY + 1000`, and `CastingConverter` to `VERY_HIGH_PRIORITY`
* Adds down the middle tests to ensure basic converters are matched appropriately
* Fixes the `NullConverter` to handle both `null` source and destination
* Removes redundant functionality from `DefaultConverter`

Question:
* Do we want the `NullConverter` to accept `null` for destination `Class` or `Type`?

I simply made the `NullConverter` do what its javadoc says it should do. Previously, it did not support `null` destination parameters and instead all requests with `null` destination parameters fell to the `DefaultConverter` which handled them. 

Below is just a list of conversions that should be handled by the `DefaultConverter` and some of the oddities I noticed with it.
* Object to Array
   * While it could handle SciJava Arrays to primitive arrays, it shouldn't since there are converters for that which are a higher priority
* Object to parameterized Collection
  * Provided the destination `Type` was an interface/abstract class which could be cast to `List` or `Set`, or it was a concrete `Type`
  * The below returns `null`, since `Collection<Number>` is an interface but cannot be cast to `List` or `Set`
```java
@SuppressWarnings("unused")
private Collection<Number> collection;

@Test
public void testDefaultConverterObjectToCollection() {
	final Field destField = ClassUtils.getField(getClass(), "collection");
	final Type destType = GenericUtils.getFieldType(destField, getClass());
	final Converter<?, ?> oc = convertService.getHandler(new Object(),
		destType);
	assertNull(oc);
}
``` 
* Number to Number
  * Though it will only handle the narrowing conversions, the `NumberToNumberConverter` handles widening conversions
* Object to String
* String to Character
* String to Enum
* Objects where the destination Class has a constructor which takes that Object

I added a short version of this list to the javadoc of `DefaultConverter`, to clarify what the `DefaultConverter` actually supports.

Please feel free to let me know if you have any questions/concerns regarding these changes, or if any additional changes are needed.

Thanks, 

Alison

### DIFF
--- a/src/main/java/org/scijava/convert/CastingConverter.java
+++ b/src/main/java/org/scijava/convert/CastingConverter.java
@@ -42,7 +42,7 @@ import org.scijava.util.GenericUtils;
  *
  * @author Mark Hiner
  */
-@Plugin(type = Converter.class, priority = Priority.FIRST)
+@Plugin(type = Converter.class, priority = Priority.EXTREMELY_HIGH)
 public class CastingConverter extends AbstractConverter<Object, Object> {
 
 	@SuppressWarnings("deprecation")
--- a/src/main/java/org/scijava/convert/DefaultConverter.java
+++ b/src/main/java/org/scijava/convert/DefaultConverter.java
@@ -80,15 +80,8 @@ public class DefaultConverter extends AbstractConverter<Object, Object> {
 
 	@Override
 	public <T> T convert(final Object src, final Class<T> dest) {
-		if (dest == null) return null;
-		if (src == null) return ConversionUtils.getNullValue(dest);
-
 		// ensure type is well-behaved, rather than a primitive type
 		final Class<T> saneDest = ConversionUtils.getNonprimitiveType(dest);
-
-		// cast the existing object, if possible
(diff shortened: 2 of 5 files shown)


ctrueden:
I rejiggered the `Priority` constants so that FIRST and LAST are not actually infinite anymore. (This was possible backwards-compatibly because the old constants with `_PRIORITY` suffix still exist but deprecated). I also added EXTREMELY_HIGH and EXTREMELY_LOW. And then I rewrote your history here to make use of all that goodness.

ctrueden:
> Do we want the `NullConverter` to accept `null` for destination `Class` or `Type`?

For now it's fine. Later, we may want things to be stricter. It makes no sense to convert either _to_ or _from_ null types. It's typically not the _type_ which is null, but the object reference.

ctrueden:
Thanks!!

PR #285, opened 2017-07-21
Add file list converters

imagejan:
This PR builds on top of https://github.com/scijava/scijava-common/pull/276.

It will allow to correctly persist values from module inputs of type `File[]` that are stored in a comma-separated (and optionally quoted) `String` values.

Note that this introduces a new dependency on Google Guava.


ctrueden:
Superseded by #286.

Issue #279, opened 2017-07-18
Use newer gentyref dependency


ghisvail:
The code for `gentyref` was moved to [GitHub](https://github.com/coekie/gentyref) and the organisation was changed accordingly, from `com.googlecode.gentyref` to `com.coekie.gentyref`. Please consider updating `scijava-common` to use the newer version, which will improve its packageability by Linux distributions.

Unfortunately, the `eventbus` dependency is a trickier one to deal with. Does anyone know where the source code repository is?

Cheers,

imagejan:
@ghisvail wrote:
> Unfortunately, the eventbus dependency is a trickier one to deal with. Does anyone know where the source code repository is?

It seems like at least a fork of this project is here:

https://github.com/wumpz/eventbus

ghisvail:
@imagejan Thanks for the pointers. Just found the official version:

https://github.com/michaelbushe/EventBus

ctrueden:
Regarding GenTyRef, SciJava Common will stop depending on it soon; see the [generic-types](https://github.com/scijava/scijava-common/compare/generic-types) branch (6be18586c201ced86c714aa138b885d8f175c24a specifically). I'll close this issue when that branch merges to master.

If it would help with packageability, we could do something similar with EventBus, since it is also a small library, and does not seem to be actively developed anymore. Please let us know if so, and I can investigate; otherwise, I'll assume you are able to package eventbus successfully.

ghisvail:
> Regarding GenTyRef, SciJava Common will stop depending on it soon

Ok, so I'd probably wait then. Right now, I have `parsington` and `imglib2` ready for submission and was walking down the dependency tree for ImageJ2, which lead me to `scijava-common`.

> If it would help with packageability, we could do something similar with EventBus

If both the `gentyref` and `eventbus` dependencies can be dropped, then that would make `scijava-common` easier to package once `parsington` is accepted. So, yes please :-)

ctrueden:
> So, yes please :-)

Great, I filed issue #282 to remind me to do that. It will likely be a few weeks, due to other priorities, but I think merging the eventbus code into SJC is a good move regardless.

[...]

PR #286, opened 2017-07-27
Improved file inputs for scripts

imagejan:
This PR supersedes https://github.com/scijava/scijava-common/pull/276 and https://github.com/scijava/scijava-common/pull/285. (I created a new one because I made further changes to the `chooseFiles` method signature that can be discussed here.)

Together with the changes in https://github.com/scijava/scijava-ui-swing/compare/master...imagejan:file-input-enhanced and https://github.com/imagej/imagej-legacy/compare/master...imagejan:file-input-enhanced, this renders `File[]` inputs possible to be resolved interactively:

* Single inputs of type `File[]` will open a file chooser that allows multiple files/directories to be opened.
* When combined with other script parameters, a list widget is created, e.g. the following Groovy script will show the input dialog shown below:

  ```groovy
  #@ File[] (label = "Input files", style = "files,extensions:tif/png/jpg") fileList
  #@ String (label = "Other parameter") stringInput

  fileList.each {
  	println "$it is being processed."
  }
  ```

  image

* A simple filter for file extensions can be created via `style` attributes as shown in the example (keyword `extensions` followed by `:` and the desired extensions separated by `/`).
* Both `File` and `File[]` input widgets are now drag-and-drop targets.
* Double-clicking an item on the file list removes it from the list.

@ctrueden if you happen to have a look, feel free to force push changes to the `file-input-enhanced` branch on my fork (I checked *Allow edits from maintainers*, so you should be able to push to my fork?!). I'll be mostly offline next week, but would love to hear/read your opinion when I'm back in the office on August 7.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>org.scijava</groupId>
 		<artifactId>pom-scijava</artifactId>
-		<version>13.1.0</version>
+		<version>17.0.0</version>
 		<relativePath />
 	</parent>
 
--- a/src/main/java/org/scijava/convert/FileListConverters.java
+++ b/src/main/java/org/scijava/convert/FileListConverters.java
@@ -1,0 +1,162 @@
+/*
+ * #%L
+ * SciJava Common shared library for SciJava software.
+ * %%
+ * Copyright (C) 2009 - 2017 Board of Regents of the University of
+ * Wisconsin-Madison, Broad Institute of MIT and Harvard, Max Planck
+ * Institute of Molecular Cell Biology and Genetics, University of
+ * Konstanz, and KNIME GmbH.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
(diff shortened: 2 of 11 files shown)


ctrueden:
Thanks @imagejan. Two issues I see:

1. We need to avoid the Guava dependency here for splitting the string.
2. The "Refactor multiple file input" commit should explicitly mention that it is OK to break the `chooseFiles` API here because that API is new and has not yet been released.

Working on it now!

imagejan:
@ctrueden wrote:
> We need to avoid the Guava dependency here for splitting the string.

I was speculating on what you said [here](https://github.com/scijava/scijava-common/issues/279#issuecomment-316425009):
> And it is likely we'll have a dependency on com.google.guava:guava in the future too.

... as I found this super useful 🙂

ctrueden:
> it is likely we'll have a dependency on com.google.guava:guava in the future

Yeah, I am torn. I meant it more generally: the ImageJ software stack will almost certainly depend on guava for, e.g., `scijava-types`, which will be a dependency of `scijava-ops`. But I am still reticent to make `scijava-common` itself depend on guava. We have minimal dependencies now, and even less or none in the future (we could bake in `eventbus` for example, and push the `parsington` dependency downstream somewhere). If SJC itself depends on guava, this will be a dealbreaker for some use cases.

ctrueden:
After a pair programming session with @imagejan, this is ready! Hooray! 👍

PR #290, opened 2017-08-28
Add Read and Write buffers

gab1one:
This PR adds two new Handles which act as readbuffer respectively writebuffer.
- *SparseBufferedHandle* is a read buffer which uses a fixed set of pages to speed up read access to an underlying handle
- *WriteBufferHandle* is a write buffer, which uses on fixed size byte array to speed up write access to an underlying handle


ctrueden:
After discussion, we will develop this logic in SCIFIO first. Once it is battle tested, we will migrate upstream to SJC here. Closing for now, until classes are fully mature.

PR #291, opened 2017-09-06
Add missing licence header to FileListPreprocessor and FileListWidget

gab1one:
Just discovered that these two where missing a header


imagejan:
Thanks for that, @gab1one!
I should remember to use that license header script before submitting PRs in the future. Where is that again?

ctrueden:
> Where is that again?

Nowhere; you run:
```
mvn license:update-file-header
```
I guess we could make a script in scijava-scripts for it, just for convenience?

> is there a plan to implement checking for license headers in scijava-maven-plugin ?

What is the goal? To fail the build if license headers are missing? We could write an enforcer rule for that, and include it in the scijava-maven-plugin next to the other enforcer rule.

PR #292, opened 2017-09-07
Add Read and Write buffer handles to scijava

gab1one:


ctrueden:
As discussed, we will look into merging this in conjunction with a use case that needs it. As an example, as SCIFIO is migrated to the new DataHandle API, we will want this PR so that we efficiently read from TIFFs remotely.

pom.xml
```diff
@@ -189,6 +189,11 @@ Konstanz, and KNIME GmbH.</license.copyrightOwners>
 			<artifactId>junit</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.mockito</groupId>
+			<artifactId>mockito-core</artifactId>
+			<version>2.9.0</version>
```
imagejan: This should be `<scope>test</scope>`, no?

gab1one: Yes, thank you 👍

gab1one:
ping @ctrueden

PR #293, opened 2017-09-21
ConsoleService: initialize MultiplePrintStream correctly, add bypass

maarzt:
The initialization of MultiplePrintStream in DefaultConsoleService
was not synchronized correctly. The global resources System.out and
System.err where modified like this: read, modify, write. This can
fail if two DefaultConsoleServices (in two scijava contexts) are
initialized in parallel.

The problem is solved in ListenableSystemStreams by using singletons.

### DIFF
--- a/src/main/java/org/scijava/console/DefaultConsoleService.java
+++ b/src/main/java/org/scijava/console/DefaultConsoleService.java
@@ -33,7 +33,6 @@
 package org.scijava.console;
 
 import java.io.OutputStream;
-import java.io.PrintStream;
 import java.util.ArrayList;
 import java.util.LinkedList;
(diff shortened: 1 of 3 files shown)


ctrueden:
Thank you for trying to fix this. But I am afraid that I really do not understand how this works. There are no comments or documentation in the code explaining it. I do not understand the purpose of the bypass thing, either. Could you please write some guidance into the code so that people looking at this are able to glean its purpose?

ctrueden:
@maarzt Is this PR still necessary? If so, would you be able to document the code as requested?

maarzt:
I don't need this anymore, so let's close the PR

PR #300, opened 2017-09-22
Fix NullPointerException when calling LogMessage::attachments()

maarzt:
Fix the following bug: 

Calling `LogMessage::attachments()` raises a `NullPointerException`, if `LogMessage::attach(Object object)` has not been called before.


ctrueden:
Thanks!

PR #306, opened 2017-11-07
Add copy method for handles

gab1one:
Adds a new method to the DataHandles class that allows for easy copying using an internal buffer

### DIFF
--- a/src/main/java/org/scijava/download/DefaultDownloadService.java
+++ b/src/main/java/org/scijava/download/DefaultDownloadService.java
@@ -37,6 +37,7 @@ import java.util.Date;
 
 import org.scijava.io.handle.DataHandle;
 import org.scijava.io.handle.DataHandleService;
+import org.scijava.io.handle.DataHandles;
 import org.scijava.io.location.Location;
 import org.scijava.plugin.Parameter;
 import org.scijava.plugin.Plugin;
@@ -70,7 +71,7 @@ public class DefaultDownloadService extends AbstractService implements
 						destination))
 			{
 				task.setStatusMessage("Downloading " + source.getURI());
-				copy(task, in, out);
+				DataHandles.copy(in, out, task);
 			}
 			catch (final IOException exc) {
 				// TODO: Improve error handling:
@@ -102,13 +103,13 @@ public class DefaultDownloadService extends AbstractService implements
 				if (isCachedHandleValid(source, cache, sourceHandle, cachedHandle)) {
 					// The data is cached; download from the cached source instead.
 					task.setStatusMessage("Retrieving " + source.getURI());
-					copy(task, cachedHandle, destHandle);
+					DataHandles.copy(cachedHandle, destHandle, task);
 				}
 				else {
 					// Data is not yet cached; write to the destination _and_ the cache.
 					task.setStatusMessage("Downloading + caching " + source.getURI());
-					copy(task, sourceHandle, //
-						new MultiWriteHandle(cachedHandle, destHandle));
+					DataHandles.copy(sourceHandle, //
+						new MultiWriteHandle(cachedHandle, destHandle), task);
 				}
 			}
 			catch (final IOException exc) {
@@ -121,31 +122,6 @@ public class DefaultDownloadService extends AbstractService implements
 	}
 
 	// -- Helper methods --
-
-	private void copy(final Task task, final DataHandle<Location> in,
-		final DataHandle<Location> out) throws IOException
-	{
-		long length;
-		try {
-			length = in.length();
-		}
-		catch (final IOException exc) {
-			// Assume unknown length.
-			length = 0;
-		}
-		if (length > 0) task.setProgressMaximum(length);
-
-		final int chunkSize = 64 * 1024; // TODO: Make size configurable.
-		final byte[] buf = new byte[chunkSize];
-		while (true) {
-			if (task.isCanceled()) return;
-			final int r = in.read(buf);
-			if (r <= 0) break; // EOF
-			if (task.isCanceled()) return;
-			out.write(buf, 0, r);
-			if (length > 0) task.setProgressValue(task.getProgressValue() + r);
-		}
-	}
 
 	private boolean isCachedHandleValid(final Location source,
 		final LocationCache cache, final DataHandle<Location> sourceHandle,
--- a/src/main/java/org/scijava/io/handle/DataHandles.java
+++ b/src/main/java/org/scijava/io/handle/DataHandles.java
@@ -9,13 +9,13 @@
  * %%
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
- * 
+ *
  * 1. Redistributions of source code must retain the above copyright notice,
  *    this list of conditions and the following disclaimer.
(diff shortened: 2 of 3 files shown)


src/test/java/org/scijava/io/handle/DataHandlesTest.java
```diff
@@ -0,0 +1,135 @@
...
+		handles = ctx.service(DataHandleService.class);
+		threadService = ctx.service(ThreadService.class);
+		eventService = ctx.service(EventService.class);
+	}
+
+	@Before
+	public void setUp() throws IOException {
+
+		final String prefix = getClass().getName();
+		inFile = File.createTempFile(prefix, "testCopyIn");
```
ctrueden: Let's use byte array handles, instead of files. No need to waste disk I/O for this one.

src/test/java/org/scijava/io/handle/DataHandlesTest.java
```diff
@@ -0,0 +1,135 @@
...
+		final Context ctx = new Context(DataHandleService.class,
+			ThreadService.class, EventService.class);
+		handles = ctx.service(DataHandleService.class);
+		threadService = ctx.service(ThreadService.class);
+		eventService = ctx.service(EventService.class);
+	}
+
+	@Before
+	public void setUp() throws IOException {
+
```
ctrueden: Kill this blank line.

src/test/java/org/scijava/io/handle/DataHandlesTest.java
```diff
@@ -0,0 +1,135 @@
...
+		final byte[] result = FileUtils.readFile(outFile);
+
+		byte[] expected = new byte[sliceSize];
+		System.arraycopy(data, 0, expected, 0, sliceSize);
+		assertArrayEquals(expected, result);
+	}
+
+	private byte[] randomBytes(final long seed) {
+		final MersenneTwisterFast r = new MersenneTwisterFast(seed);
+		final byte[] ldata = new byte[2938740];
```
ctrueden: You use underscore in numeric constant above, but not here. Let's be consistent!

src/test/java/org/scijava/io/handle/DataHandlesTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+	private static DataHandleService handles;
+	private File inFile;
+	private File outFile;
+	private byte[] data;
+	private static ThreadService threadService;
+	private static EventService eventService;
+
+	@BeforeClass
+	public static void classSetup() {
```
ctrueden: Let's make this non-static. I.e.: spin a new context in `@Before` rather than using `@BeforeClass` and potentially bleeding state between tests.

src/test/java/org/scijava/io/handle/DataHandlesTest.java
```diff
@@ -0,0 +1,135 @@
+
```
ctrueden: Missing license header.

src/main/java/org/scijava/download/DefaultDownloadService.java
```diff
@@ -70,7 +71,7 @@ public Download download(final Location source, final Location destination) {
 						destination))
 			{
 				task.setStatusMessage("Downloading " + source.getURI());
-				copy(task, in, out);
+				DataHandles.copy(in, out, task);
```
ctrueden: :+1: :+1:

src/main/java/org/scijava/io/handle/DataHandles.java
```diff
@@ -38,9 +38,12 @@
...
 import java.lang.reflect.Method;
 
+import org.scijava.io.location.Location;
+import org.scijava.task.Task;
+
 /**
  * Utility methods for working with {@link DataHandle}s.
- * 
+ *
  * @author Curtis Rueden
```
ctrueden: Add yourself as another author.

src/main/java/org/scijava/io/handle/DataHandles.java
```diff
@@ -108,4 +111,135 @@ private static synchronized void initUTFMethod() {
...
 	}
+
+	/**
+	 * Copies all bytes from the input to the output handle. Reading and writing
+	 * start at the current positions of the handles.
+	 *
+	 * @param in the input handle
+	 * @param out the output handle
+	 * @return the number of bytes copied
+	 * @throws IOException
```
ctrueden: Need to write _when_ the exception is thrown.

src/main/java/org/scijava/io/handle/DataHandles.java
```diff
@@ -108,4 +111,135 @@ private static synchronized void initUTFMethod() {
...
+	}
+
+	/**
+	 * Copies up to <code>length</code> bytes from the input to the output handle.
+	 * Reading and writing start at the current positions of the handles. Stops
+	 * early if there are no more bytes available from the input handle.
+	 *
+	 * @param in the input handle
+	 * @param out the output handle
+	 * @param length the number of bytes to copy
```
ctrueden: Should state here that `0` means all bytes, same as in the other signature.

src/main/java/org/scijava/io/handle/DataHandles.java
```diff
@@ -108,4 +111,135 @@ private static synchronized void initUTFMethod() {
...
+	/**
+	 * Copies up to <code>length</code> bytes from the input to the output handle,
+	 * reporting the progress to the provided task. Reading and writing start at
+	 * the current positions of the handles. Stops early if there are no more
+	 * bytes available from the input handle. Uses a buffer of the provided size,
+	 * instead of using the default size.
+	 *
+	 * @param in input handle
+	 * @param out the output handle
+	 * @param length maximum number of bytes to copy, will copy all bytes if set
```
ctrueden: More grammatically correct would be `;` instead of `,`. Otherwise it's a [comma splice](https://en.wikipedia.org/wiki/Comma_splice).

src/main/java/org/scijava/io/handle/DataHandles.java
```diff
@@ -108,4 +111,135 @@ private static synchronized void initUTFMethod() {
...
+	 * @return the number of bytes copied
+	 * @throws IOException
+	 */
+	public static long copy(final DataHandle<Location> in,
+		final DataHandle<Location> out, final long length, final Task task,
+		final int bufferSize) throws IOException
+	{
+
+		// get length of input
+		long inputlength;
```
ctrueden: You can achieve `final` on this variable if you structure this section a little more carefully.

src/main/java/org/scijava/io/handle/DataHandles.java
```diff
@@ -108,4 +111,135 @@ private static synchronized void initUTFMethod() {
...
+				r = in.read(buffer, 0, remaining);
+			}
+			else {
+				r = in.read(buffer);
+			}
+			if (r <= 0) break; // EOF
+			if (task != null && task.isCanceled()) break;
+			out.write(buffer, 0, r);
+			totalRead += r;
+			if (task != null) task.setProgressValue(task
```
ctrueden: Should use curly braces for the consequent here, so formatter does the right thing.

gab1one:
I updated the pr with all the changes you requested.

ctrueden:
Thanks!!

PR #307, opened 2017-11-17
FileHandle: don't unnecessarily init a RandomAccessFile on close

gab1one:
If we close a handle on a file that does not exist, we do not need to
initialize the RandomAccessFile, just to close it again. This fixes the problem of
empty files appearing when closing a FileHandle to a file that does not exist.


gab1one:
I don't think a single handle should be shared between threads at all, as this can very easily result in data corruption unless the threads are coordinating their access somehow.

gab1one:
@ctrueden  with the added synchronized access do you mean something like I did in commit fdfc4ef ?
That should protect us from the scenario with the closing of a FileHandle that is just initializing.

ctrueden:
@gab1one Let's just make the `close()` method synchronized. Simpler, less buggy, no less efficient. No?

PR #309, opened 2017-12-04
Logging, shorten console output, fix listeners

maarzt:
This fixes to issues:

* Don't output too much meta information about log messages on the console
* AbstractLogService called the method notifyListeners in the wrong place, which might cause every LogListener to be called twice per log message.


ctrueden:
Thank you!

PR #312, opened 2018-01-09
[For discussion] Add fall-back handle when we can not resolve a handle.

gab1one:
@ctrueden and I talked about supporting "Location-only" use cases, where no `Handle` for a `Location` is needed, as no real data access is happening. One example is the `TestImgLocation` in Scifio, that creates virtual images based on the location information, we will probably need something similar for ImageJ-Omero.

The problem is that during the parsing process, scifio will try to open a handle on such a location, resulting in a `NullPointerException` if we do not provide a HandleImplementation. Instead of creating a dedicated `DummyHandle` for each such `Location` type we can make use of one central fallback like the one here.

### DIFF
--- a/src/main/java/org/scijava/io/handle/FallBackHandle.java
+++ b/src/main/java/org/scijava/io/handle/FallBackHandle.java
@@ -1,0 +1,89 @@
+
+package org.scijava.io.handle;
+
+import java.io.IOException;
+
+import org.scijava.Priority;
+import org.scijava.io.location.Location;
+import org.scijava.plugin.Plugin;
+
+/**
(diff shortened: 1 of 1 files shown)


gab1one:
Per discussion with Curtis we found a better approach to handle this issue

ctrueden:
Just to very briefly describe what we decided: we are going to change the `Parser` API so that it does not have a hard dependency on a `DataHandle` being present for the `Location`. It should be the responsibility of individual concrete parser implementations to request a `DataHandle` when they need such a thing. And for the corresponding `Checker` of the format to reject locations without a `DataHandle` being available. The default/abstract parser layer does not, and should not, need a `DataHandle` for anything.

PR #313, opened 2018-01-10
Fix NullPointerException in LoggerPreprocessor

imagejan:
This handles the case of `loggerName` being `null` when running scripts that contain a `#@ Logger` parameter.


maarzt:
:+1: Thanks for the fix. It's perfectly fine and ready to be merged.

PR #315, opened 2018-01-29
BytesLocation: improve javadoc

gab1one:
- Adds note about size of handles created using BytesLocation(int).
- Adds explanation of BytesLocation(byte[], int, int)


ctrueden:
Perfect! Sorry I dropped the ball till now.

PR #321, opened 2018-05-06
LocationService#resolve() add fall-back to FileLocation

gab1one:
Before giving up on parsing a given string, we try interpreting it as a
local file. This allows us to parse local paths, without requiring
users to transform them to URIs first. The current implementation will only resolve existing files but we could change that easily.

### DIFF
--- a/src/main/java/org/scijava/io/location/LocationService.java
+++ b/src/main/java/org/scijava/io/location/LocationService.java
@@ -32,6 +32,7 @@
 
 package org.scijava.io.location;
 
+import java.io.File;
 import java.net.URI;
 import java.net.URISyntaxException;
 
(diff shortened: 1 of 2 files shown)


ctrueden:
This feels hacky. Is there some reason that `LocationResolver` does not simply also have a `resolve(String)`, upon which we can lean when `LocationService#resolve(String)` is called? That would be much more straightforward, elegant and extensible.

gab1one:
The `LocationResolver` plugins are registered and located via the `scheme` part of the URI, how can we find the matching one if we do not have one?

gab1one:
I reworked the code to be cleaner and added a test.

gab1one:
I think I found a nicer way to solve this which also does not force the resolution as a FileLocation. Instead the missing `scheme` is set to `file://` and the normal resolution code path is taken.

PR #323, opened 2018-05-06
Add test if a handle exists

gab1one:


src/main/java/org/scijava/io/handle/DataHandleService.java
```diff
@@ -63,4 +65,18 @@
...
+
+	/**
+	 * Convenience method to test whether it describes an existing file.
+	 *
+	 * @param Location the location to test
+	 * @return The result of {@link DataHandle#exists()} on a newly created handle
+	 *         on this location
+	 * @throws IOException
+	 */
+	default boolean handleExists(Location Location) throws IOException {
```
ctrueden: Please just call it `exists`. The name `handleExists` implies to me the question "does a `DataHandle` exist?" which is not what we are going for. And make the variable name `location` small case.

gab1one: done :+1:

PR #330, opened 2018-06-19
Fix #329: Empty RAF creation on read

### LINKED ISSUE #329: FileHandle creates an empty file when trying to read from non-existing file
If we try to read from a file that does not exists, the FileHandle currently creates this file silently, resulting in an 0 byte large file appearing on disk, this is clearly unwanted behavior. Instead the reading should fail with either an IOException or (depending on the called method) return a fitting default value.

gab1one:
When trying to read from a non-existing file, the FileHandle creates a RandomAccessFile, which results in an empty file appearing on disk. This PR fixes this by throwing an IOException when trying to read from a non-existent file.

### DIFF
--- a/src/main/java/org/scijava/io/handle/FileHandle.java
+++ b/src/main/java/org/scijava/io/handle/FileHandle.java
@@ -60,9 +60,12 @@ public class FileHandle extends AbstractDataHandle<FileLocation> {
 
 	// -- FileHandle methods --
 
-	/** Gets the random access file object backing this FileHandle. */
+	/**
+	 * Gets the random access file object backing this FileHandle. If the
+	 * underlying file does not exist yet, it will be created.
+	 */
 	public RandomAccessFile getRandomAccessFile() throws IOException {
-		return raf();
+		return writer();
 	}
 
 	public String getMode() {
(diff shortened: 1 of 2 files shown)


src/test/java/org/scijava/io/handle/FileHandleTest.java
```diff
@@ -108,4 +110,30 @@ public void testNotCreatedByClose() throws IOException {
...
+		assertFalse(nonExistentFile.exists());
+
+		final FileLocation loc = new FileLocation(uri);
+		final DataHandle<?> handle = dhs.create(loc);
+		assertFalse(handle.exists());
+		try {
+			handle.read(); // this will fail as there is no underlying file!
+			fail("Read successfully from non-existing file!");
+		}
+		catch (final Exception exc) {
```
imagejan: How about catching only `IOException` instead of `Exception` here?

gab1one: Good point, I just pushed an updated version that does just that.

src/test/java/org/scijava/io/handle/FileHandleTest.java
```diff
@@ -108,4 +110,30 @@ public void testNotCreatedByClose() throws IOException {
...
+		final DataHandleService dhs = ctx.service(DataHandleService.class);
+
+		final File nonExistentFile = //
+			File.createTempFile("FileHandleTest", "none xistent file");
+		final URI uri = nonExistentFile.toURI();
+		assertTrue(nonExistentFile.delete());
+		assertFalse(nonExistentFile.exists());
+
+		final FileLocation loc = new FileLocation(uri);
+		final DataHandle<?> handle = dhs.create(loc);
```
ctrueden: We should use try-with-resources here, no?
```
try (final DataHandle<?> handle = dhs.create(loc)) {
  ...
}
```
So that it definitely gets closed at the end, regardless of the outcome of the test

src/main/java/org/scijava/io/handle/FileHandle.java
```diff
@@ -101,199 +101,199 @@ public Date lastModified() {
...
 	@Override
 	public long offset() throws IOException {
-		return raf().getFilePointer();
+		return exists() ? writeRaf().getFilePointer() : 0;
 	}
 
 	@Override
 	public long length() throws IOException {
-		return exists() ? raf().length() : -1;
+		return exists() ? writeRaf().length() : -1;
```
ctrueden: Conceptually, this should be `readRaf()` since querying the length is a "read" operation. (I realize that in practice, it won't matter due to the `exists()` check beforehand, but the code looks strange/wrong with `writeRaf()` here.)

src/main/java/org/scijava/io/handle/FileHandle.java
```diff
@@ -306,11 +306,38 @@ public synchronized void close() throws IOException {
...
+	 * if the underlying file exists on disk. This prevents accidental creation of
+	 * empty RAF file when calling read operations on a non-existent file.
+	 *
+	 * @return the internal {@link RandomAccessFile} or creates if if needed
+	 * @throws IOException if the {@link RandomAccessFile} could not be created,
+	 *           or the backing file does not exists.
+	 */
+	private RandomAccessFile readRaf() throws IOException {
+		if (raf == null) {
+			if (!exists()) {
```
ctrueden: I would prefer this `exists()` check live in the `initRAF()` method. It can be done by making the method `initRAF(final boolean create)` and then writing `if (!create && !exists()) throw ...`. Then, here in `readRAF()`, simply put `if (raf == null) initRAF(false);`—and correspondingly, `initRAF(true)` in the `writeRAF()` method. Then the two methods will be symmetrical, and IMHO easier to understand.

Note also the case of `RAF`—it is an acronym, so should be all capitals, not camel case `Raf`. Maybe we want to call them `reader()` and `writer()` instead of `readRaf()` and `writeRaf()`? I feel like `readRaf()` and `writeRaf()` are somehow too long/unwieldy.

gab1one: Great suggestion, this makes the code much cleaner :)

gab1one:
Thanks for the input @ctrueden I integrated your comments and it should be good to go now.

src/main/java/org/scijava/io/handle/FileHandle.java
```diff
@@ -306,12 +306,46 @@ public synchronized void close() throws IOException {
...
+	 * @throws IOException if the {@link RandomAccessFile} could not be created,
+	 *           or the backing file does not exists.
+	 */
+	private RandomAccessFile reader() throws IOException {
+		if (raf == null) initRAF(false);
+		return raf;
+	}
+
+	/**
+	 * Initializes the {@link RandomAccessFile},
```
ctrueden: `,` ➡️ `.`

src/main/java/org/scijava/io/handle/FileHandle.java
```diff
@@ -306,12 +306,46 @@ public synchronized void close() throws IOException {
...
+	 *
+	 * @param create whether to create the {@link RandomAccessFile} if the
+	 *          underlying file does not exist yet.
+	 * @throws IOException if the {@link RandomAccessFile} could not be created,
+	 *           or the backing file does not exists and the {@code create}
```
ctrueden: `exists` ➡️ `exist`

src/main/java/org/scijava/io/handle/FileHandle.java
```diff
@@ -306,12 +306,46 @@ public synchronized void close() throws IOException {
...
+	 *           parameter was set to {@code true}.
```
ctrueden: `true` ➡️ `false`

gab1one:
> Just a handful of little things now!
Also included :)

ctrueden:
Thanks @gab1one !!

PR #331, opened 2018-06-26
When processing #@script, register module with ModuleService

imagejan:
Also set the priority of ScriptDirectiveScriptProcessor to high to allow processing directives before other script parameters.


ctrueden:
Thanks!

PR #332, opened 2018-07-03
SingletonService: update instances when available plugins change

gab1one:
Previously instances of  SingletonService did not update their instance mapping when pluings got added or removed. Now they do :tada:

### DIFF
--- a/src/main/java/org/scijava/plugin/AbstractSingletonService.java
+++ b/src/main/java/org/scijava/plugin/AbstractSingletonService.java
@@ -32,13 +32,17 @@
 
 package org.scijava.plugin;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.scijava.event.EventHandler;
 import org.scijava.log.LogService;
 import org.scijava.object.ObjectService;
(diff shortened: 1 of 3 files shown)


src/main/java/org/scijava/plugin/AbstractSingletonService.java
```diff
@@ -84,20 +85,63 @@ public ObjectService objectService() {
...
+		}
+	}
+
+	@EventHandler
+	protected void onEvent(final PluginsAddedEvent event) {
+		if (instanceMap == null) return;
+		// collect singleton plugins
+		final List<PluginInfo<PT>> singletons = new ArrayList<>();
+		for (final PluginInfo<?> pluginInfo : event.getItems()) {
+			if (SingletonPlugin.class.isAssignableFrom(pluginInfo.getPluginType())) {
```
ctrueden: I believe this should be `getPluginType().isAssignableFrom(pluginInfo.getPluginType())`.

Could you please fix it, and then update the unit test to prove that when we add plugins of multiple types, only the correct ones are picked up by the respective services?

gab1one: I added a test for this and it was indeed a problem, so I fixed it as you suggested.

ctrueden:
Thank you @gab1one !

PR #335, opened 2018-08-17
Fix FileHandleTest on Windows

imagejan:
* Use try-with-resources on DataHandle in FileHandleTest
* Revert commit f436b7d (Work around Windows file deletion problems)


ctrueden:
Thank you very much!

PR #336, opened 2018-09-04
DataHandleService: better handling of failure states

gab1one:
No longer throws a NullPointerException when a handle can not be created


ctrueden:
Please add a commit that documents the pre- and post-conditions of these methods, especially when `null` is given as the argument. This PR changes the assumptions, which is fine, but as long as we are thinking about nulls, let's be explicit about the intended behavior now.

PR #337, opened 2018-09-28
Datahandle test improvements and Endianness fixes

gab1one:
The coverage of DataHandle by DataHandleTest was not sufficient, a lot of methods where untested. This PR greatly extends the test coverage and fixes some discovered bugs.

### DIFF
--- a/src/main/java/org/scijava/io/handle/DataHandle.java
+++ b/src/main/java/org/scijava/io/handle/DataHandle.java
@@ -563,7 +563,7 @@ public interface DataHandle<L extends Location> extends WrapperPlugin<L>,
 	default String readLine() throws IOException {
 		// NB: Adapted from java.io.RandomAccessFile.readLine().
 
-		final StringBuffer input = new StringBuffer();
+		final StringBuilder input = new StringBuilder();
 		int c = -1;
 		boolean eol = false;
 
--- a/src/main/java/org/scijava/io/handle/DataHandle.java
+++ b/src/main/java/org/scijava/io/handle/DataHandle.java
@@ -250,6 +250,11 @@ public interface DataHandle<L extends Location> extends WrapperPlugin<L>,
 	/** Sets the native encoding of the stream. */
 	void setEncoding(String encoding);
 
+	/**
+	 * @return a 8 byte long buffer array used for type conversions
+	 */
+	byte[] conversionBuffer();
+
 	/** Reads a string of arbitrary length, terminated by a null char. */
 	default String readCString() throws IOException {
 		final String line = findString("\0");
@@ -515,7 +520,7 @@ public interface DataHandle<L extends Location> extends WrapperPlugin<L>,
 
 	@Override
 	default short readShort() throws IOException {
-		final byte[] buf = new byte[2];
+		final byte[] buf = conversionBuffer();
 		final int read = read(buf, 0, 2);
 		if (read < 2) throw new EOFException();
 		return Bytes.toShort(buf, isLittleEndian());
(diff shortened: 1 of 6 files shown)


src/main/java/org/scijava/io/handle/DataHandle.java
```diff
@@ -514,18 +515,10 @@ default int readUnsignedByte() throws IOException {
...
-			ch0 = read();
-			ch1 = read();
-		}
-		else {
-			ch1 = read();
-			ch0 = read();
-		}
-		if ((ch0 | ch1) < 0) throw new EOFException();
-		return (short) ((ch0 << 8) + (ch1 << 0));
+		final byte[] buf = new byte[2];
```
ctrueden: This will destroy performance. We cannot allocate a new byte array with every short read. You could use an instance field buffer of sufficient size to avoid the issue.

src/main/java/org/scijava/io/handle/DataHandle.java
```diff
@@ -540,68 +533,20 @@ default char readChar() throws IOException {
...
-		}
-		else {
-			ch3 = read();
-			ch2 = read();
-			ch1 = read();
-			ch0 = read();
-		}
-		if ((ch0 | ch1 | ch2 | ch3) < 0) throw new EOFException();
-		return ((ch0 << 24) + (ch1 << 16) + (ch2 << 8) + (ch3 << 0));
+		final byte[] buf = new byte[4];
```
ctrueden: Same: performance will tank. Need a small reusable `byte[]` as instance field. This should be OK, since these methods are already not thread-safe.

src/main/java/org/scijava/io/handle/DataHandle.java
```diff
@@ -540,68 +533,20 @@ default char readChar() throws IOException {
...
-			ch6 = read();
-			ch5 = read();
-			ch4 = read();
-			ch3 = read();
-			ch2 = read();
-			ch1 = read();
-			ch0 = read();
-		}
-		if ((ch0 | ch1 | ch2 | ch3 | ch4 | ch5 | ch6 | ch7) < 0) {
+		byte[] buf = new byte[8];
```
ctrueden: Need to use a reusable buffer.

src/main/java/org/scijava/io/handle/DataHandle.java
gab1one: How can we cache this in the interface though? Should we require all DataHandle implementations to provide a buffer array which is used for conversions?

gab1one:
Hi @ctrueden, I addressed your performance concern in https://github.com/scijava/scijava-common/pull/337/commits/7b7e03bd5b322a859521026d3d7678194b126170
I am not sure I like this solution though, it might be better to just push the affected methods into AbstractDataHandle and not expose the `conversionBuffer` in the interface.

ctrueden:
@gab1one I considered different solutions to address the issue of `conversionBuffer()` as public API in the interface.

One idea I had was to use a static `ThreadLocal<byte[]>` in a utility class, to give back a unique `byte[]` per thread. I benchmarked it, and the performance is equivalent to your direct field storing the `byte[]` per instance of `DataHandle`.

Then I thought: where will this utility class live? What will it be called? It is pretty specialized, always returning a `byte[8]`. It could be generalized, but maybe [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) applies here. I figured we want this to be an internal implementation detail, so the utility class itself should maybe be package-protected.

But then I wondered: is it ever useful to be able to override the conversion buffer access? I am torn. There might be cases where it is helpful downstream; I don't know. (But this also might be a case of YAGNI.)

I am definitely not eager to move so many default method implementations in the abstract layer, since it makes implementing the interface directly much more complicated, even though as you point out it would certainly have better encapsulation regarding this issue.

In the absence of a clearly superior alternative, I'm going to merge this as-is with the `conversionBuffer()` method exposed in the public API.

PR #338, opened 2018-10-04
SingletonServiceTest: extend with tests from #271.

gab1one:
This adds the additional tests from #271 to the SingletonServiceTest


ctrueden:
Thank you @gab1one and @stelfrich!

PR #340, opened 2018-10-15
Add stream locations

gab1one:
These are currently used for Compressed input streams #274 and https://github.com/scijava/scijava-io-http


gab1one:
Thanks for fixing this, will add the Test ASAP

gab1one:
I am actually not sure if we need the `DefaultBufferedStreamHandle` at all, I implemented a test and discovered a series of issues with it's design. I think the `ReadBufferDataHandle` / `WriteBufferDataHandle` are better and fulfill the same role.

PR #346, opened 2018-12-14
ReadBufferDataHandle: ensure pages are fully read

gab1one:
Reading bytes with `DataHandle#read(bytes[], off, len)` results in *up
to* `len` bytes being read into the provided array.
In ReadBufferHandleDataHandle the number of read bytes was not checked,
this lead to partially filled pages resulting in read errors. Now the
reading is repeated until the page is either full or EOF is reached.


ctrueden:
Thank you for fixing this! :+1:

PR #349, opened 2019-01-23
BytesLocation: Add method to set the name for the Location

gab1one:
In some instances libarary users want to create a BytesLocation that has
a specific name, e.g. so that it is correctly detected by a Service that
performs a name based lookup.

This came up as a requirement several times during the fiji hackathon at it4inovations


ctrueden:
Thanks @gab1one.

I see that `getName()` already existed in the `Location` API, so I'm OK with adding the constructors etc. to `BytesLocation`. I do have some concerns about using the name for technical things like differentiation of content types, as we discussed via voice the other day. Might be good to discuss further on image.sc or on a GitHub issue here.

Issue #333, opened 2018-07-19
Handle spaces in parameter style declarations


imagejan:
Currently, these two parameter declarations behave differently:

    #@ File[] (style="files,extensions:xls/xlsx") files

and

    #@ File[] (style="files, extensions:xls/xlsx") files

The latter ignores the file extension filter because after splitting, the style string has a leading space.

Let's fix this by changing `style.split(",")` to `style.trim().split("\\s*,\\s*")` in all places where it is relevant:

* [ ] `scijava-common`:
https://github.com/scijava/scijava-common/blob/821c8294f753535c86a23dad9e277f78d572a79b/src/main/java/org/scijava/widget/DefaultWidgetModel.java#L133

* [ ] `scijava-ui-swing`:
https://github.com/scijava/scijava-ui-swing/blob/d0802b1a09d380898f77022b4010682cf330fc28/src/main/java/org/scijava/ui/swing/widget/SwingFileWidget.java#L210

* [ ] `imagej-legacy`:
https://github.com/imagej/imagej-legacy/blob/325fd0933778dabd6ee8c09867961c698ffe945c/src/main/java/net/imagej/legacy/plugin/MacroRecorderPostprocessor.java#L95

ctrueden:
Let's make a utility method of WidgetStyle to handle this in a unified way?

imagejan:
@ctrueden do we want to be case-insensitive when testing `style` attributes (i.e. allowing all of `message`, `MESSAGE` and `mEsSaGe`)?

PR #354, opened 2019-05-13
Safeguard against null style arg in UserInterface#chooseFile

imagejan:
This fixes the issue reported by @kephale [on the forum](https://forum.image.sc/t/error-for-file-type-imagej2-parameter/25585?u=imagejan), where the absence of a `style` attribute on a `File` parameter leads to a `NullPointerException`.

This PR is just a quick fix, the method should probably be rewritten when addressing #333.

### DIFF
--- a/src/main/java/org/scijava/ui/UserInterface.java
+++ b/src/main/java/org/scijava/ui/UserInterface.java
@@ -162,9 +162,12 @@ public interface UserInterface extends RichPlugin, Disposable {
 	 */
 	default File chooseFile(final File file, final String style) {
 		final String title;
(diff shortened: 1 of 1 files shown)


ctrueden:
Thanks @imagejan. I am concerned that it is possible for `style` to ever be null. But certainly, being defensive here is a good thing.

imagejan:
@ctrueden wrote:
> I am concerned that it is possible for `style` to ever be null.

From `DefaultWidgetModel`, it seems that `style` being null is allowed/considered:

https://github.com/scijava/scijava-common/blob/821c8294f753535c86a23dad9e277f78d572a79b/src/main/java/org/scijava/widget/DefaultWidgetModel.java#L129-L137

PR #360, opened 2019-10-14
LogMessage: Do not print null message

frauzufall:
* Prevent printing of log message if message is null

How to reproduce:
```
DefaultLogger log = new DefaultLogger(System.out::println, LogSource.newRoot(), LogLevel.INFO);
try {
	((Object)null).equals(null);
} catch (NullPointerException e) {
	log.error(e);
}
```
.. prints:
```
[ERROR] null
java.lang.NullPointerException
	at org.scijava.log.LogTest.run(LogTest.java:12)
```

With this PR, it prints
```
[ERROR] java.lang.NullPointerException
	at org.scijava.log.LogTest.run(LogTest.java:12)
```


ctrueden:
Thank you!

PR #361, opened 2019-10-14
Gateway: implements Disposable

frauzufall:
* make `Gateway` implement `Disposable`
* delegate `dispose()` to `context` by default

Reason: It took me a while to figure out I have to call `ij.context().dispose()` to avoid memory leaks. This would make it a bit easier to find. It was also mentioned by @ctrueden in [this forum thread](https://forum.image.sc/t/how-to-find-memory-leaks-in-plugins-what-needs-to-get-disposed/10211/14?u=frauzufall).

Could this change have any implications which might not be covered by existing tests?


ctrueden:
Thanks @frauzufall !

PR #366, opened 2019-11-08
Add abstract delegate converter class

### LINKED ISSUE #182: Create delegate converter abstract class
We don't want the framework to chain conversion, but it is useful to be able to create delegate converters that explicitly rely on a known reliable chained conversion. We should standardize this with an an abstract class.

imagejan:
This PR closes #182 and adds `AbstractDelegateConverter` that allows to specify new converters that chain together two subsequent conversions like this:

```java
@Plugin(type = Converter.class)
public class StringToDatasetConverter extends
	AbstractDelegateConverter<String, File, Dataset>
{

	@Override
	public Class<Dataset> getOutputType() {
		return Dataset.class;
	}

	@Override
	public Class<String> getInputType() {
		return String.class;
	}

	@Override
	protected Class<File> getDelegateType() {
		return File.class;
	}
}
```

<s>I am not sure about how to implement the test: is it possible to create private nested classes that can be discovered as `Plugin`s during test time?</s>

<s>I'd appreciate some advice here if you have time, @ctrueden.</s>

**EDIT**: I think I figured it out. I force-pushed the latest commit, adding a test for the abstract class. This should now be ready for review.

Test coverage of the added classes: 100% 🙂

### DIFF
--- a/src/main/java/org/scijava/convert/AbstractDelegateConverter.java
+++ b/src/main/java/org/scijava/convert/AbstractDelegateConverter.java
@@ -1,0 +1,29 @@
+
+package org.scijava.convert;
+
+import org.scijava.plugin.Parameter;
+
+/**
+ * Abstract superclass for {@link Converter} plugins that delegate to other
+ * converters to chain two conversion steps together.
+ * 
+ * @author Jan Eglinger
+ * @param <I> the input type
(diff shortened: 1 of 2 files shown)


ctrueden:
Cool, thanks a lot @imagejan!

Issue #48, opened 2014-05-01
Handle case of choosing among multiple display plugins for a single Dataset


ctrueden:
When a Dataset can be displayed in multiple ways must be able to handle which one is run as cleanly as possible. We should be able to set each display plugin's priority and run the Display with the lowest priority. If there is a collision in priorities between two or more display plugins then run another plugin that allows user to choose which display plugin to run.

Migrated-From: http://trac.imagej.net/ticket/391

PR #368, opened 2019-12-03
DefaultDisplayService: Choose matching display with highest priority

frauzufall:
I ran into issue https://github.com/scijava/scijava-common/issues/48 and though this PR does not give the user a chance to choose which display to use, it at least makes the `DefaultDisplayService` choose the one with the highest priority. 

I also added a test, but could not yet test it myself, because I get a bunch of unrelated errors with this repo which I have to investigate. But if anyone already has an opinion on this in the meantime or can try it out, let me know.

### DIFF
--- a/src/main/java/org/scijava/display/DefaultDisplayService.java
+++ b/src/main/java/org/scijava/display/DefaultDisplayService.java
@@ -33,9 +33,11 @@
 package org.scijava.display;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.LinkedList;
 import java.util.List;
 
+import org.scijava.Prioritized;
 import org.scijava.display.event.DisplayActivatedEvent;
 import org.scijava.display.event.DisplayCreatedEvent;
 import org.scijava.display.event.DisplayDeletedEvent;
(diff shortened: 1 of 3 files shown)


frauzufall:
Mh, my test it not great because all tests now have this dummy class `CustomTextDisplay` in scope. How would a better test look like?

imagejan:
@frauzufall wrote:
> all tests now have this dummy class CustomTextDisplay in scope

I had a similar issue in https://github.com/scijava/scijava-table/pull/8 where I added a `FakeTableIOPlugin` that's on the classpath for all tests (see https://github.com/scijava/scijava-table/pull/8/files). I tried first having this as a nested inner class, but that didn't work. If there's a better solution, I'd be interested as well 🙂

src/test/java/org/scijava/display/DisplayTest.java
```diff
@@ -132,4 +132,14 @@ public void testText() {
...
 
+	@Test
+	public void testMultipleDisplaysPriorityMatch() {
+		final Context context = new Context(DisplayService.class);
+		final DisplayService displayService =
+				context.getService(DisplayService.class);
+		final String name = "Text";
+		final String value = "Hello";
+		Display<?> display = displayService.createDisplay(name, value);
+		assertEquals(CustomTextDisplay.class, display.getClass());
```
imagejan: In addition to asserting that `CustomTextDisplay` gets chosen, can we also test that we actually do get *multiple* candidate `Display`s? Otherwise the name of this method would be misleading, no?

ctrueden:
> How would a better test look like?

I recently made it easier to create `PluginInfo` objects easily; see f31a468bc3c17cf133ce1545fbe1b4da872facfb. Here is an example of how this could be used in tests:

```java
Context ctx = new Context(...); // whatever services you want
PluginInfo<Display> info = PluginInfo.create(CustomTextDisplay.class, Display.class);
info.setPriority(Priority.HIGH);
ctx.service(PluginService.class).addPlugin(info);
... now do the actual test ...
```

Typically the context creation would be in the `@Before` method, and context disposal in the `@After` method, but if you have only a single test that wants the `CustomTextDisplay` in the plugin index, you could add it manually like that at the beginning of that test. Or if all tests of a particular class want it, then make a `@Before` method for that class adding it there.

frauzufall:
Thanks for the feedback! I used the strategy  @ctrueden described and adjusted the test along @imagejan's note. The tests are running now and my use case is also working with it.

src/main/java/org/scijava/display/DefaultDisplayService.java
```diff
@@ -219,20 +221,28 @@ public boolean isUniqueName(final String name) {
 
 	@Override
 	public Display<?> createDisplayQuietly(final Object o) {
+		final List<Display<?>> matchingDisplays = getMatchingDisplays(o);
+		if(matchingDisplays.size() > 0) {
+			// use the display with the highest priority
+			Display<?> display = matchingDisplays.stream().max(Comparator.comparing(Prioritized::getPriority)).get();
```
ctrueden: I don't think you need this. The displays should already always be returned from the `PluginService` in high-to-low priority. Did you observe otherwise?

frauzufall:
Hi @ctrueden, you are right, the PluginService returns the displays in the correct order, which means this PR is not needed at all. I did not properly check if the test I added was actually failing before (reminded me of our discussion about whether the PR related test should be the first commit or the last one :upside_down_face: ).. I also cannot reproduce my use case where it failed. Probably put the priority on the wrong class.... Next time I start with an issue.

Sorry for the trouble! I still learned something from your comments :+1:

imagejan:
We could still add the test for multiple displays to test if the highest priority wins. I think the current `DisplayTest` doesn't test this, right?

PR #377, opened 2020-01-09
Respect autoFill parameter in script parameters

### LINKED ISSUE #376: Parameter autoFill attribute not respected in scripts
Consider the following script:

```groovy
#@ Dataset (autoFill=false) img
println img
```

While I would expect a dialog with a dropdown choice to appear here, the parameter is always auto-filled (by `ActiveDatasetPreprocessor` from `imagej-common`, but it contains logic similar to `org.scijava.module.process.AbstractSingleInputPreprocessor`, so I'm filing it here).

The same issue with parameters types `Img`, `ImagePlus` etc., and tested also with `persist=false` and `required=false`.

We need to check if `item.isAutoFill()` really returns `false` in these cases. I don't think there's a test for this specific attribute yet, is there?

imagejan:
This adds a new method `setAutoFill(boolean)` to `MutableModuleItem` and fixes `ParameterScriptProcessor` to set the `autoFill` value according to the parameter attribute.

`ScriptInfoTest` is updated with a parameter testing `autoFill=false` that fails without the changes in this commit.

Fixes #376.

### DIFF
--- a/src/main/java/org/scijava/module/DefaultMutableModuleItem.java
+++ b/src/main/java/org/scijava/module/DefaultMutableModuleItem.java
@@ -54,6 +54,7 @@ public class DefaultMutableModuleItem<T> extends AbstractModuleItem<T>
 	private final Type genericType;
 	private ItemIO ioType;
 	private ItemVisibility visibility;
+	private boolean autoFill;
 	private boolean required;
 	private boolean persisted;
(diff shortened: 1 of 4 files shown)


ctrueden:
Waah! Great catch, @imagejan! And even fixing my typos! 🏆

Issue #382, opened 2020-04-21
Support for file chooser that works with both files and directories


kephale:
I'm looking into a file chooser that works for both Files and Directories (e.g. should work with tif and n5 datasets which are directories). i presume the place to introduce this change is https://github.com/scijava/scijava-common/blob/448a8bd51c4910834e0a1f0e7af393acf3e7d7a5/src/main/java/org/scijava/ui/UserInterface.java#L163

imagejan:
Both files and directories would be stored in a `File` object on the Java side.

The `FileListWidget` already allows for the `both` style:

https://github.com/scijava/scijava-common/blob/560ad964a7a1e859edd4dbe4fc48d7d9e09523f2/src/main/java/org/scijava/widget/FileListWidget.java#L51-L56

To allow the same for single files, we need to add a new style to `FileWidget` as well, and then adapt the `UserInterface#chooseFile` method as you pointed out.

LauLauThom:
> To allow the same for single files, we need to add a new style to FileWidget as well

Is it really necessary to create a new style ?  
I dont understand why we could not use the `both` style from the `FileListWidget` for the `FileWidget` too. Is it important that a given style is part of the class it is used in ?
Or can the new `FileWidget` style also encode for `both` ?

I had the same interrogation for https://github.com/scijava/scijava-common/issues/197#issuecomment-619216086. Using only `directory` for both `FileWidget` and `FileListWidget` so we dont have to document an additional `directories` style.

imagejan:
Well, keeping the constants in the classes where they're used would be consistent. As `FileWidget` defined `OPEN`, `SAVE` (both of which are "files-only") and `DIRECTORY`, this was not sufficient for the requirements of `FileListWidget` where we wanted to select files, directories or both, but never save. That's why originally we defined new styles there.
But I agree, we could also unify this and keep all styles in the `FileWidget` class. Maybe even be more concise and say "open-file", "open-directory", "open-both" and "save-file". But now we also have to ensure backwards-compatibility 🙁 so if we define new styles or change them, we'll have to keep the others for a while (forever?) as deprecated...

@ctrueden any opinion/suggestion?

LauLauThom:
Alright, maybe the better is the enemy of the good here :P 
I think we should introduce the least changes.
So as a sum up this is the current situation

__File styles__
- "open"
- "save"
- "directory"
- "both" (missing) -> new one ? or use FileList.both ?

__FileList styles__
- "files"
- "directories"
[...]

Issue #197, opened 2015-10-15
Add possibility to filter by file type in FileWidget


imagejan:
It would be great if you could ask for a specified file type in a parameter definition:

```
// @File(label="Open file", style="open", type="*.xls;*.xlsx") inputFile
```

so that the resulting FileWidget dialog would filter the directory by file type.

ctrueden:
Adding more attributes to the `@Parameter` annotation in Java land is not really tenable for stuff like this. However, it is a great fit for the existing `style` attribute. The point of `style` is to provide some "hints" for how you want the input widget to work. So that seems like a good match here.

Maybe we could do something like:

```java
// @File(label="Open file", style="open:*.xls;*.xlsx") inputFile
```

And extend the mechanism as needed (maybe by additional plugins...) to support whatever syntax is useful for specifying file filters. This could include useful stuff like "image formats only" (in all formats ImageJ supports), "text formats only", etc.

The downside of tying this metadata to the `style` is that it would not _enforce_ those limitations at validation time. Whereas with things like `min` and `max` and `stepSize` those requirements could be enforced beyond the widget level.

Validation in general could certainly use some improvement in the framework; see scijava/scijava-common#47 for details.

imagejan:
In particular, it would be useful if you can provide a `java.io.FileFilter` or `java.io.FileNameFilter`. For scripting in particular, we might provide shortcuts for `org.apache.commons.io.filefilter.WildcardFilter` and `org.apache.commons.io.filefilter.RegexFileFilter`.

Do we have to change `FileWidget` to account for this? Or rather parse the `style` annotation, split it at `:` and then create a filter?

For example:
`style="open:*.xls;*.xlsx"` => `new WildcardFilter({"*.xls", "*.xlsx"})`
`style="open:/.*\.(xls|xlsx)/` => `new RegexFileFilter({"*.xls", "*.xlsx"})`

----

Also, we should consider whether this filter restricts the allowed files, or we use it simply for facilitating file choice by offering multiple filter types in the "File of Type" dropdown for `JFileChooser`.

imagejan:
With https://github.com/scijava/scijava-ui-swing/commit/73281ecafabaef8e4eee2a75bda339e90cf86507, filtering by file extension is now possible with the following syntax:

```
#@ File (style = "open, extensions:xls/xlsx") inputFile
```

So this is implemented at the UI level. If at some point we want to restrict file types at the `scijava-common` level, we should re-open this issue.

[...]

PR #384, opened 2020-04-30
Add WidgetStyle utility class to centralize style logic

imagejan:
This pull request addresses #333 and is a first step toward fixing #197 and #382.

The methods in the new `WidgetStyle` utility class should replace `isStyle()` implementations and other case logic around (possibly multiple) style attributes of parameters in the following places:

* [ ] `scijava-ui-swing`:
  * [ ] `chooseFile` and `chooseFiles` in [`AbstractSwingUI`](https://github.com/scijava/scijava-ui-swing/blob/ac3cbb189749a3fc55d30034ba4a2f8270dcd069/src/main/java/org/scijava/ui/swing/AbstractSwingUI.java#L136-L213)
  * [ ] `createFileFilter` in [`SwingFileWidget`](https://github.com/scijava/scijava-ui-swing/blob/d0802b1a09d380898f77022b4010682cf330fc28/src/main/java/org/scijava/ui/swing/widget/SwingFileWidget.java#L195-L240)
  * [ ] some helper methods in [`SwingChoiceRadioWidget`](https://github.com/scijava/scijava-ui-swing/blob/557d79230d0f70eaae8573e242f0a02c66b07d62/src/main/java/org/scijava/ui/swing/widget/SwingChoiceRadioWidget.java#L112-L126)

* [ ] `batch-processor`:
  * [ ] `getTargetWidgetStyle` in [`FileBatchInputProvider`](https://github.com/scijava/batch-processor/blob/d5e1b3c33880014cdc907f9ebce452c795eec3f3/src/main/java/org/scijava/batch/input/FileBatchInputProvider.java#L65-L85)

* [ ] `imagej-plugins-batch`:
  * [ ] `getTargetWidgetStyle` in [`DatasetBatchInputProvider`](https://github.com/imagej/imagej-plugins-batch/blob/2d0ad6f3c87adb6bce7bc19ee9cea3f8c97e20ba/src/main/java/net/imagej/batch/input/DatasetBatchInputProvider.java#L92-L108)

* [ ] here in [`imagej-legacy`](https://github.com/imagej/imagej-legacy/blob/325fd0933778dabd6ee8c09867961c698ffe945c/src/main/java/net/imagej/legacy/plugin/MacroRecorderPostprocessor.java#L93-L98), and finally

* [ ] some places in [`scijava-ui-awt`](https://github.com/scijava/scijava-ui-awt/blob/aabb0f917069543a6f96b93c983db7cdec39932c/src/main/java/org/scijava/ui/awt/widget/AWTFileWidget.java#L110), and [three](https://github.com/scijava/scijava-ui-pivot/blob/9069fda303852da3c4a3d127b47e667350dbf391/src/main/java/org/scijava/ui/pivot/widget/PivotFileWidget.java#L103) [times](https://github.com/scijava/scijava-ui-pivot/blob/9069fda303852da3c4a3d127b47e667350dbf391/src/main/java/org/scijava/ui/pivot/widget/PivotNumberScrollBarWidget.java#L90) in [`scijava-ui-pivot`](https://github.com/scijava/scijava-ui-pivot/blob/9069fda303852da3c4a3d127b47e667350dbf391/src/main/java/org/scijava/ui/pivot/widget/PivotNumberSliderWidget.java#L88)

---

Remaining questions:

* I wasn't sure about the naming of `getStyleModifiers`. The intended use case is for styles like `"extensions:tiff/tif"` to get the list of allowed extensions, but there might be more general use cases.

* Should `getStyleModifiers` return an array of strings directly from `split("/")`? Or rather a `Set<String>` with trimmed, lower-case values?

### DIFF
--- a/src/main/java/org/scijava/widget/DefaultWidgetModel.java
+++ b/src/main/java/org/scijava/widget/DefaultWidgetModel.java
@@ -129,12 +129,7 @@ public class DefaultWidgetModel extends AbstractContextual implements WidgetMode
 
 	@Override
 	public boolean isStyle(final String style) {
-		final String widgetStyle = getItem().getWidgetStyle();
-		if (widgetStyle == null) return style == null;
-		for (final String s : widgetStyle.split(",")) {
-			if (s.equals(style)) return true;
-		}
-		return false;
+		return WidgetStyle.isStyle(getItem(), style);
 	}
 
 	@Override
--- a/src/main/java/org/scijava/widget/WidgetStyle.java
+++ b/src/main/java/org/scijava/widget/WidgetStyle.java
@@ -1,0 +1,35 @@
+package org.scijava.widget;
+
+import org.scijava.module.ModuleItem;
+
+public class WidgetStyle {
+	private WidgetStyle() {
+		// prevent instantiation of utility class
+	}
+
+	public static boolean isStyle(String widgetStyle, String target) {
+		if (widgetStyle == null || target == null)
+			return widgetStyle == target;
+		for (final String s : widgetStyle.split(",")) {
+			if (s.trim().toLowerCase().equals(target.toLowerCase())) return true;
+		}
+		return false;
+	}
+
+	public static boolean isStyle(ModuleItem<?> item, String target) {
+		return isStyle(item.getWidgetStyle(), target);
+	}
+
+	public static String[] getStyleModifiers(String widgetStyle, String target) {
+		if (widgetStyle == null || target == null)
+			return null;
+		String[] styles = widgetStyle.split(",");
+		for (String s : styles) {
+			if (s.trim().toLowerCase().startsWith(target.toLowerCase())) {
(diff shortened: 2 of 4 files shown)


imagejan:
Thanks @ctrueden for merging this.
In the light of https://github.com/scijava/scijava-ui-swing/pull/52, I now think we should maybe replace/modify `getStyleModifiers` to do something a bit more general, i.e. return everything after the `:` (`String` instead of `String[]`) and leave it to the caller to split the returned string (e.g. in the case of multiple extensions, such as `extensions:tiff/tif`) or to use the returned string entirely (e.g. with `format:#0.000`).

Now that `2.85.0` is released, we probably have to keep the API for backwards compatibility, right? : )

ctrueden:
> Now that 2.85.0 is released, we probably have to keep the API for backwards compatibility, right? : )

If we change it immediately, I'd be OK with breaking it. It's not part of pom-scijava et al. yet.

PR #385, opened 2020-05-25
DefaultRecentFileService: add dispose method

frauzufall:
I ran this loop when working on memory issues in `SCIFIO`:
```
for (int i = 0; i < 1000; i++) {
	SCIFIO scifio = new SCIFIO(new Context());
	scifio.dispose();
}
```
It was very slow in comparison to creating the Context with specific services. I narrowed it down to `new Context(RecentFileService.class))` and this patch seems to make the `RecentFileService` clean up properly on dispose.


ctrueden:
Good catch!

PR #386, opened 2020-05-28
AbstractInputHarvester: make getObjects() not return duplicates

frauzufall:
* both the `convertService` and the `objectService` attach objects of a
given type to the result list of `AbstractInputHarvester:getObjects`
* this results into the same objects being in the list twice
* using a set avoids duplicates in the list

Example for testing:
```
@Plugin(type = Command.class)
public class TestCommand implements Command {

    @Parameter
    private Img input;

    @Parameter
    private Img input2;

    @Override
    public void run() {}

    public static void main(final String... args) throws Exception {
        final ImageJ ij = new ImageJ();
        ij.ui().showUI();
        Object img = ij.io().open("/home/random/blobs.tif");
        ij.ui().show(img);
        ij.command().run(TestCommand.class, true);
    }
}
```

### DIFF
--- a/src/main/java/org/scijava/widget/AbstractInputHarvester.java
+++ b/src/main/java/org/scijava/widget/AbstractInputHarvester.java
@@ -30,7 +30,9 @@
 package org.scijava.widget;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import org.scijava.AbstractContextual;
 import org.scijava.convert.ConvertService;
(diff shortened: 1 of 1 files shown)


imagejan:
Thanks @frauzufall for investigating!

This might be part of a solution for https://github.com/imagej/imagej-legacy/issues/230.

frauzufall:
@ctrueden this might help fix macro recorder issues though I had no time to confirm this yet. But since it's only a small change fixing an issue which multiple people have told me about, it would be great to merge this.

ctrueden:
Awesome @frauzufall! This has been a loooongstanding bug. Thanks for fixing it! And sorry I missed it till now.

PR #390, opened 2020-06-19
ScriptService: add getScript(String,String) signature

### LINKED ISSUE #389: Add convenience method to retrieve ScriptInfo/ModuleInfo from script
As suggested by @ctrueden [on gitter](https://gitter.im/scijava/scijava-common?at=5ee92a6724a3382d5d6eded0), we should add a convenience method to `ScriptService` that returns a `ScriptInfo` or `ModuleInfo` for a given script.

Something like `getScriptInfo(String filename, String script)` ?

imagejan:
This makes it easier to retrieve a `ScriptInfo` object from a script string (without going via `File`).

Closes #389.

### DIFF
--- a/src/main/java/org/scijava/script/DefaultScriptService.java
+++ b/src/main/java/org/scijava/script/DefaultScriptService.java
@@ -30,6 +30,7 @@
 package org.scijava.script;
 
 import java.io.File;
+import java.io.StringReader;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


imagejan:
I refactored `ScriptServiceTest` a bit to initialize the `Context` in a `@Before` method, and dispose it `@After` the tests.

@ctrueden what do you think about the re-use of the `ScriptLanguage` from `ScriptInfoTest$BindingSizes`? Should we instead make `BindingSizes` its own class within the tests? There are actually *two* nested `ScriptLanguage` implementations in the tests currently: [`ScriptInfoTest$BindingSizes`](https://github.com/scijava/scijava-common/blob/73743b0000bd1682f29cff4005d5414abc5736c7/src/test/java/org/scijava/script/ScriptInfoTest.java#L390) and [`ScriptEngineTest$Rot13`](https://github.com/scijava/scijava-common/blob/73743b0000bd1682f29cff4005d5414abc5736c7/src/test/java/org/scijava/script/ScriptEngineTest.java#L128).

PR #394, opened 2020-08-11
DefaultThreadService: allow faster JVM shutdown

maarzt:
I noticed that even a simple program as the following will keep the JVM alive for several seconds:
```java
public class Main {
     public static void main(String... args) {
           new Context();
     }
}
```
This is caused by threads created by the DefaultThreadService, which run for a view seconds even after all the tasks are completed.

This PR fixes the late JVM shutdown. The threads created by DefaultThreadService are made deamon threads and will no longer block the JVM shutdown. This unfortunately causes the JVM to shutdown before all tasks submitted to ThreadService are completed. This is fixed by adding a LifeCounter, that counts the tasks submitted to ThreadService and keeps the JVM alive until all tasks are completed.

### DIFF
--- a/src/main/java/org/scijava/thread/DefaultThreadService.java
+++ b/src/main/java/org/scijava/thread/DefaultThreadService.java
@@ -73,7 +73,16 @@ public final class DefaultThreadService extends AbstractService implements
 
 	private boolean disposed;
 
+	private LifeCounter lifeCounter = new LifeCounter();
+
 	// -- ThreadService methods --
+
+	public DefaultThreadService() {
+		// NB: We need to call EventQueue.isDispatchThread() once before the JVM
+		// shuts down. Otherwise EventQueue.isDispatchThread won't work as
+		// expected during JVM shutdown.
+		EventQueue.isDispatchThread();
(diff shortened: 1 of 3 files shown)


hinerm:
@maarzt @ctrueden I don't think this PR is necessary as-is because `Context` creation has an implicit expectation of calling `Context#dispose` when ready to shut down, in which case the process quickly exits.

But I do think we could do more to make this expectation clearer. One option would be javadoc, but another thought was having `Context` implement `AutoCloseable` with a trivial implementation:
```
@Override
public void close() throws Exception {
	dispose();
}
```

which would facilitate simple use cases:
```
public static void main(String... args) throws Exception {
	try (Context context = new Context()) { }
}
```

and, if not closed, will trigger `Resource leak` warnings in Eclipse. What do you think?

maarzt:
This PR breaks the behavior of System.exit() by adding the blocking shutdown hook. Thank @tpietzsch for pointing that out.

maarzt:
I'm in favor of @hinerm's suggestion to make Context implement AutoClosable.

PR #399, opened 2020-10-16
Added format attribute to @Parameter

BoudewijnvanLangerak:
Added format attribute to @Parameter for formating the value in the SwingNumberWidget

Intended to fix:
https://github.com/scijava/scijava-ui-swing/issues/45

### DIFF
--- a/src/main/java/org/scijava/module/ModuleItem.java
+++ b/src/main/java/org/scijava/module/ModuleItem.java
@@ -205,4 +205,6 @@ public interface ModuleItem<T> extends BasicDetails {
 	/** Sets the item's current value with respect to the given module. */
 	void setValue(Module module, T value);
 
+	/** Gets the item's format*/
+    String getFormat();
 }
--- a/src/test/java/org/scijava/command/ValidNumberCommandTest.java
+++ b/src/test/java/org/scijava/command/ValidNumberCommandTest.java
@@ -50,7 +50,7 @@ import org.scijava.plugin.Plugin;
  * 
  * @author Curtis Rueden
  */
-public class InvalidCommandTest {
+public class ValidNumberCommandTest {
 
 	private CommandService commandService;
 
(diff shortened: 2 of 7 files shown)


src/main/java/org/scijava/module/ModuleItem.java
```diff
@@ -205,4 +205,6 @@
 	/** Sets the item's current value with respect to the given module. */
 	void setValue(Module module, T value);
 
+	/** Gets the item's format*/
+    String getFormat();
```
imagejan, line 209: This new API will break all implementations of `ModuleItem` (except those extending `AbstractModuleItem`, of course). It requires increasing the major version of `scijava-common` to `3.x.x.`.

src/test/java/org/scijava/command/ValidNumberCommandTest.java
```diff
@@ -0,0 +1,101 @@
...
+import org.scijava.ItemIO;
+import org.scijava.ValidityProblem;
+import org.scijava.plugin.Parameter;
+import org.scijava.plugin.Plugin;
+
+/**
+ * Test commands for verifying that invalid module parameters are dealt with
+ * using proper error handling.
+ * 
+ * @author Curtis Rueden
```
imagejan, line 51: @ctrueden isn't the author of this class, right? I suppose this was just copy-pasted from the other test class?

imagejan (commented):
Thanks @BoudewijnvanLangerak for working on this. Unfortunately, your PR includes significant changes to the public API, and I doubt it will find its way into SciJava any time soon. @ctrueden might want to include something like this into his work on SJC3?

In the past, Curtis suggested to use and extend the `style` attribute for similar functionality (e.g. like `style="files,extensions:tiff"` for `File[]` parameters).

We could for example do something like this:

```
#@ Double (style="slider,format:0.000000")
```

imagejan:
I'm closing this is now in favor of a backwards-compatible solution using the `style` parameter attribute, as implemented in https://github.com/scijava/scijava-ui-swing/pull/54.

PR #403, opened 2020-10-25
Always ignore commented script parameters

### LINKED ISSUE #398: Inconsistent behavior with commented script parameters
When `#@` script parameters are commented out (with `//` or `#`, dependent on the language used), the behavior differs depending on the position of the parameter definitions in the script (due to the support of both old-style and new-style parameters).

Consider the following Groovy scripts (all working as expected):

```groovy
/*
 * Some comment as header
 * (This script uses new-style parameter syntax)
 */
#@ String text
```

```groovy
#@ String text
/*
 * (This script uses a new-style parameter, but at the beginning of the script)
 */
```

```groovy
// @String text
/*
 * (Old-style parameters will only be parsed at the very beginning of the code)
 */
```

Now, in the scripts using the new-style (`#@`) syntax, if you comment out the script parameter with `//`, the first script will stop showing the dialog, while the second script will *still* prompt for the `text` parameter:

```groovy
/*
 * The following parameter declaration will be ignored
 * because # is not at the beginning of the line
 */
// #@ String text
```

```groovy
// #@ String text
/*
 * ^^^ !
 * The above parameter matches the old-style syntax, even if both // and # are present
 */
```

We can probably change this to respect (and ignore) comments if and only if the script is using new-style syntax, by changing how we match the line when parsing old-style parameters here:

https://github.com/scijava/scijava-common/blob/3d64e2a2ea9f6eab5d75c7e778dc3da75bd03f90/src/main/java/org/scijava/script/process/ParameterScriptProcessor.java#L143

The question is if that would break any scripts in the wild that rely on the current behavior. There's definitely [some scripts](https://github.com/NEUBIAS/neubias-springer-book-2021/blob/0354e1d7a5f84706e2537bc66dc93b3d36204276/Ch01_Batch_Processing_Methods_in_ImageJ/code/02a_Process_Folder_Path.ijm#L1-L11) already that rely on *commenting* script parameters, but it would probably be ok to restore consistent behavior if the commented parameter occurs at the very beginning of the script.

@ctrueden what's your opinion?

NB: the behavior for Python is the same: `# #@String text` is ignored, except if in the first line.

imagejan:
Closes #398.

Before this commit, commented-out new-style (`#@`) script parameters were ignored when in the script _body_, but parsed when in the script _header_, because they matched the pattern for old-style parameter syntax.
We now explicitly omit lines with _new-style syntax preceded by non-word characters_ from being matched, so the behavior is consistent, independent of the position of the parameter in the script.

Without this change in `ParameterScriptProcessor`, the test added in this commit would fail with:

```
[ERROR]   ParameterScriptProcessorTest.testScriptParameterParsing:49 expected null, but was:<commentedHeaderParameter: required=true, persisted=true>
```

### DIFF
--- a/src/main/java/org/scijava/script/process/ParameterScriptProcessor.java
+++ b/src/main/java/org/scijava/script/process/ParameterScriptProcessor.java
@@ -129,7 +129,7 @@ public class ParameterScriptProcessor implements ScriptProcessor {
 
 	@Override
 	public String process(final String line) {
-		// parse new-style parameters starting with @# anywhere in the script.
+		// parse new-style parameters starting with #@ anywhere in the script.
 		if (line.matches("^#@.*")) {
 			final int at = line.indexOf('@');
 			return process(line, line.substring(at + 1));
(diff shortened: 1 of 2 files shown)


ctrueden:
Perfect! Thank you, @imagejan!

PR #405, opened 2020-10-30
WidgetStyle: add javadoc and new method

imagejan:
While `getStyleModifiers()` is useful for things like `"extensions:png/gif/bmp"`, this commit adds a new `getStyleModifier()` method that simply returns the suffix after the colon (`:`), for cases like `"format:#0.001"`.

See also https://github.com/scijava/scijava-ui-swing/pull/52.

### DIFF
--- a/src/main/java/org/scijava/widget/WidgetStyle.java
+++ b/src/main/java/org/scijava/widget/WidgetStyle.java
@@ -35,6 +35,16 @@ public class WidgetStyle {
 		// prevent instantiation of utility class
 	}
 
+	/**
+	 * Check whether a given widget style contains the target style.
+	 * 
+	 * @param widgetStyle
+	 *            The style declaration to test, usually a comma-separated
+	 *            {@code String}; trailing spaces are ignored.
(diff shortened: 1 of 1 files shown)


imagejan:
@ctrueden just saw your comment here: https://github.com/scijava/scijava-common/pull/384#issuecomment-719629651

If we're free to change the API of `WidgetStyle`, we could consider a) changing the return type of `getModifiers()` from `String[]` to `Set<String>` or b) removing the `getModifiers` signature entirely.

Also, I'm still open to (re-)naming suggestions :-)

imagejan:
@hinerm do you think this can be merged? Or alternatively, we can adapt the API introduced in #384, but I'm afraid there has been a release by now, so it would be a breaking API change (although no SciJava component was using it yet) ...

Anyhow, I think it would be good to get this done, so we can proceed with https://github.com/scijava/scijava-ui-swing/pull/54 and https://github.com/scijava/scijava-ui-swing/pull/56.

hinerm (approved):

Issue #410, opened 2020-11-30
IllegalStateException in logging framework when executing a script


ctrueden:
During the first session of the sciview tutorial at I2K 2020:

![sciview-IllegalStateException-I2K-2020](https://user-images.githubusercontent.com/556626/100632575-9c425780-32f2-11eb-954b-fe223869c6d2.png)

Likely a bug in the SciJava logging framework.

CC @kephale @skalarproduktraum @maarzt

imagejan:
What was the script run from the `TextEditor` to trigger this issue?

ctrueden:
```python
# @SciView sv

sv.addSphere()

print('My first script!')
```

On @kephale's Linux box.

Probably not too enlightening though... looks more like a race condition to me. Here's the relevant line:

https://github.com/scijava/script-editor/blob/script-editor-0.5.6/src/main/java/org/scijava/ui/swing/script/TextEditor.java#L2042

It's just a debug log statement during cleanup phase.

ctrueden:
Huh.

https://github.com/scijava/scijava-common/blob/scijava-common-2.85.0/src/main/java/org/scijava/log/CallingClassUtils.java#L52-L62

Maybe the `Thread.currentThread().getStackTrace()` somehow fails, due to the multithreaded stuff going on? Seems like we should fail more gracefully here rather than throwing `IllegalStateException`.

imagejan:
FWIW, I can reproduce the `ClassNotFoundException` (that presumably leads to this issue here) using the following Groovy script:

```groovy
stackTrace = Thread.currentThread().getStackTrace()

className = stackTrace[1].getClassName()
println className

clazz = Class.forName(className)
println clazz
```

which throws:

```
java.lang.ClassNotFoundException: java_lang_Thread$getStackTrace$0
```

ctrueden:
Ahh! I missed that `Class.forName` was being called. Yeah, that's not good. Thanks for catching that, @imagejan.

maarzt:
I would suggest this fix https://github.com/scijava/scijava-common/pull/411

I failed to reproduce the problem. I therefor couldn't make sure that the fix really works.

skalarproduktraum:
Thanks for the fix @maarzt! @kephale's Linux box is possessed by a mean-spirited entity that causes random errors, I suggested to him repeatedly to start fresh with it, therefore I am not totally surprised this is not reproducible 😏

maarzt:
> mean-spirited entity

[...]

PR #411, opened 2020-12-03
LogService: Make getLevel() more robust against ClassNotFoundException

maarzt:
This should fix issue: https://github.com/scijava/scijava-common/issues/410

Previously: If fine grained log levels are set LogService would call `CallingClassUtils.getCallingClass()` to find the appropriate log level for this class. This process would fail if the calling class could not be loaded with the default class loader. A IllegalStateException would be thrown instead.

This PR fixes this behavior. The class name is used rather the the class object. This avoids using `Class.forName(...)` if possible. But `Class.forName(...)` is still used to check for the presence of the `IgnoreAsCallingClass` annotation.

### DIFF
--- a/src/main/java/org/scijava/log/CallingClassUtils.java
+++ b/src/main/java/org/scijava/log/CallingClassUtils.java
@@ -43,12 +43,44 @@ public final class CallingClassUtils {
 	}
 
 	/**
+	 * Inspects the stack trace to return the name of the class that calls
+	 * this method, but ignores every class annotated with @IgnoreAsCallingClass.
+	 * <p>
+	 * If every class on the stack trace is annotated, then the class at the
+	 * root of the stack trace is returned.
+	 */
+	public static String getCallingClassName() {
+		StackTraceElement[] stackTrace = Thread.currentThread().getStackTrace();
+		for (int i = 1; i < stackTrace.length - 2; i++) {
+			String className = stackTrace[i].getClassName();
+			if (!hasIgnoreAsCallingClassAnnotation(className)) return className;
+		}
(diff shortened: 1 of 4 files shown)


src/main/java/org/scijava/log/CallingClassUtils.java
```diff
@@ -43,12 +43,44 @@ private CallingClassUtils() {
...
+		for (int i = 1; i < stackTrace.length - 2; i++) {
+			String className = stackTrace[i].getClassName();
+			if (!hasIgnoreAsCallingClassAnnotation(className)) return className;
+		}
+		return stackTrace[stackTrace.length - 1].getClassName();
+	}
+
+	private static boolean hasIgnoreAsCallingClassAnnotation(String className) {
+		try {
+			Class< ? > clazz = Class.forName(className);
```
ctrueden, line 63: Could you please instead use `Context.getClassLoader()`? We try to never use `Class.forName` directly throughout the SciJava codebase; the thread context class loader is used in many places throughout ImageJ, but it might return null, so you need a fallback. See cd67323c13678c931cb5e1e798149c45ef5b21a9.

ctrueden (changes requested):
@maarzt Thanks a bunch for working on this! Especially while on leave. Much appreciated.

I just have one request, to avoid using `Class.forName` in favor of SciJava's unified class loading access method.

src/main/java/org/scijava/log/CallingClassUtils.java
maarzt: Ok, it's done

Issue #406, opened 2020-11-01
Version 2.85.0 does not open tif images any more


sunsear:
With version 2.85.0 it is no longer possible to successfully open tif images. Larger images put the CPU at 100% for minutes, usually takes a second or so. Smaller images result in:

![Schermafbeelding 2020-11-01 om 14 27 32](https://user-images.githubusercontent.com/2803378/97804261-edf1b680-1c4e-11eb-8982-b9a6dd22aba3.png)

#### Succesful opening of images
Make a project with:
```
	<parent>
		<groupId>org.scijava</groupId>
		<artifactId>pom-scijava</artifactId>
		<version>29.2.1</version>
		<relativePath/>
	</parent>
```
It is now possible to use IJLauncher to open an image and do stuff with it.

#### Failed opening of images

1. Add the property:
`		<scijava-common.version>2.85.0</scijava-common.version>`
2. Start up the IJLauncher.
3. Open an image
4. CPU goes to 100% and image is not opened

imagejan:
Thanks @sunsear for reporting. For the record, here's what changed between `scijava-common-2.83.3` (pinned in `pom-scijava-29.2.1`) and `scijava-common-2.85.0`:

https://github.com/scijava/scijava-common/compare/scijava-common-2.83.3...scijava-common-2.85.0

I *suspect* it might be the changes to `(Default)IOService` that make use of the new `Location`-based API in SCIFIO... @ctrueden @hinerm any insights?

sunsear:
I did a little compare there, but there was so much change in IO related stuff that I couldn’t see what the recant part would be. Hope @ctrueden knows!

imagejan:
The reason is this commit added between `2.83.4` and `2.84.0`:

> Use Location, not String, in the I/O API (https://github.com/scijava/scijava-common/commit/b7285c5f807891097b5405b270952fcd32f2253c)

... and the fact that [`DatasetIOPlugin`](https://github.com/scifio/scifio/blob/bdb8ab764d58243d9f3a9f3936d468d695268a9c/src/main/java/io/scif/io/DatasetIOPlugin.java) in SCIFIO wasn't adapted to these changes yet.

sunsear:
@ctrueden , @imagejan , any progress on this? We are stuck with https://github.com/scijava/scijava-ui-swing/pull/54 because of this issue.

Look forward to hearing from you!

imagejan:
[...]

PR #418, opened 2021-02-26
Fix issue #406 without a location call handler supportsOpen(string)

karlduderstadt:
This commit fixes an issue caused by calling the Location version of the
method supportsOpen() in all cases when many IOPlugin do not yet
implement that Location version of that method and therefore always
return false. Instead now IOService getOpener(string) directly calls the
string version of supportsOpen. This ensures older IOPlugin that can
support the file provided are discovered and used.


hinerm:
@karlduderstadt would you be able to build and test with the latest SCIFIO? I just merged https://github.com/scifio/scifio/pull/451 which I assume fixes the issue in the case of images.

@ctrueden This change seems reasonable to me, because it moves the translation from `String` to `Location` to the [AbstractIOPlugin](https://github.com/scijava/scijava-common/blob/a0f04b2f6b992a737e962c9a7d8f7b5a5076c89b/src/main/java/org/scijava/io/AbstractIOPlugin.java#L55) instead of the `IOService`. But am I missing any intent here?

karlduderstadt:
@hinerm I just built and tested the latest SCIFIO with scifio/scifio#451 with scijava-common 46ca0d0 and indeed the addition of jdeschamps ensures that File>Open... now correctly opens images. 

But my custom file types still fail because I have not yet added supportsOpen(final Location source). I will certainly make those small updates to my custom IOPlugins.

Looking at the code in scijava-common, clearly the intention was to ensure backwards compatibility with any past IO Plugins that were written to work with string paths. So I still think we should merge this. Otherwise, there need to be changes that strictly enforce use of Locations. But I think it is nice to support backwards compatibility with it doesn't appear to cost much. I think this is one of those cases.

Thanks a lot for checking on this and writing back @hinerm!

PR #422, opened 2021-05-11
CommandInfo: allow shadowed Service parameters

### LINKED ISSUE #408: Allow private service fields
See https://github.com/fiji/FilamentDetector/issues/24#issuecomment-733133909.

imagejan:
When a `Command` extends another `Command`, we often have both classes use specific services.
It's better to be able to name these `Service` parameters consistently, than to avoid shadowing parameters at all costs. So let's exclude all `Service` parameters from the validity check.

Closes #408.


ctrueden:
Great!

PR #423, opened 2021-05-13
Skip FileListPreprocessor for autofill=false parameters 

imagejan:
<s>This pull request is a work in progress.</s>

This pull request changes the API of `ModuleService` and `DefaultModuleService` to include an optional parameter `boolean acrossTypes` that allows checking for solitary unresolved parameters within *all* inputs/outputs independent of their type, while maintaining the backwards-compatible method signatures.

This allows to replace some duplicated logic in helper methods of `FilePreprocessor` and `FileListPreprocessor` by a call to `moduleService.getSingleInput()`, which makes them respect an `autoFill=false` parameter annotation and avoid calling `uiService.chooseFile/chooseFiles` for those parameters.

The intended behavior is verified by two new tests that, before this PR, were failing on their last assertion.

NB: The tests added here call `uiService.setHeadless(false)` in order to set their custom default UI.

### DIFF
--- a/src/main/java/org/scijava/module/DefaultModuleService.java
+++ b/src/main/java/org/scijava/module/DefaultModuleService.java
@@ -260,27 +260,23 @@ public class DefaultModuleService extends AbstractService implements
 	}
 
 	@Override
-	public <T> ModuleItem<T> getSingleInput(final Module module,
-		final Class<T> type)
-	{
-		return getTypedSingleItem(module, type, module.getInfo().inputs());
+	public <T> ModuleItem<T> getSingleInput(final Module module, final Class<T> type, boolean acrossTypes) {
+		return getTypedSingleItem(module, type, module.getInfo().inputs(), acrossTypes);
 	}
 
 	@Override
-	public <T> ModuleItem<T> getSingleOutput(final Module module,
-		final Class<T> type)
-	{
-		return getTypedSingleItem(module, type, module.getInfo().outputs());
+	public <T> ModuleItem<T> getSingleOutput(final Module module, final Class<T> type, boolean acrossTypes) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/scijava/module/DefaultModuleService.java
```diff
@@ -478,24 +474,25 @@ private void assignInputs(final Module module,
...
 	private ModuleItem<?> getSingleItem(final Module module,
-		final Collection<Class<?>> types, final Iterable<ModuleItem<?>> items)
+		final Collection<Class<?>> types, final Iterable<ModuleItem<?>> items, boolean acrossTypes)
 	{
 		ModuleItem<?> result = null;
 
 		for (final ModuleItem<?> item : items) {
 			final String name = item.getName();
 			if (!item.isAutoFill()) continue; // skip unfillable inputs
 			if (module.isInputResolved(name)) continue; // skip resolved inputs
```
imagejan, line 494: While working on this pull request, I noticed that the above line is specific to *inputs* only, so the method probably doesn't show the intended behavior when called with *outputs*.

The current unit tests seem to miss the method signatures specific to outputs (`getSingleOutput`).

I'll open a separate issue for this.

ctrueden:
Sorry it's taken me so long to look at this. I now have a dedicated PR review day per week, and this one is on my shortlist right now to review, but I'm trying to apply the 80/20 rule to do the easiest ones first. My hangup on this one is the addition of the `acrossTypes` parameter. I know it's optional, and does not break backwards compatibility, but it does complexify the API, so I'm going to have a think on whether we truly need it for this new feature. Maybe not this week, maybe not next week, but soon, and for the rest of our lives. 🏠🇪🇸

PR #430, opened 2022-02-06
IOService recognizes a new 3rd state of loading data

xulman:
The `IOService` allows (likely among other functionalities) a client-code to be hooked into the drag&drop chain of openers. Each opener in this chain is asked if it `supportsOpen()` and `Object open(final Location source)` is called on the first opener supporting the given input.

However, the `open()` call expects data to be loaded that, in the drag&drop scenario, [is directly submitted to `uiService.show()`](https://github.com/fiji/IO/blob/3aa1c1a93abf199766c86384ab5cdb968981b841/src/main/java/HandleExtraFileTypes.java#L549-L552).

This PR attempts to solve the use case when the `open()` actually wants to use its own way of presenting the loaded data. The proposal defines a special, unique, signalling object `IOService.GOVERNING_APP_STARTED` that must be returned from the `open()` calls in such cases, that is, when 3rd party app is opened on the incoming (dropped-in) file.

This PR works in conjunction with https://github.com/fiji/IO/pull/18

### DIFF
--- a/src/main/java/org/scijava/io/IOService.java
+++ b/src/main/java/org/scijava/io/IOService.java
@@ -87,6 +87,15 @@ public interface IOService extends HandlerService<Location, IOPlugin<?>>,
 		return null;
 	}
 
+	/** A special type of "openned data" that can be returned by the
+	 * {@link #open(String)} and {@link #open(Location)} methods, and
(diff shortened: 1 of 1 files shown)


xulman:
deprecated in favour of `SelfShowableContent` -> https://github.com/scijava/scijava-common/pull/473

PR #432, opened 2022-03-29
Task interface and DefaultTask modifications - backward compatible

NicoKiaru:
With this PR task events are published automatically when an asynchronous task (task.run(runnable)))
is started and finished. Additionally, task can be run in a synchronous manner and can trigger task events
with additional methods added to the Task interface.

This commit brings additional methods to the Task interface (empty default implementations provided):
* start and finish method: a job can notify when it is starting and ending in a synchronous manner
* setCancelCallback(runnable) method: Can be used to add a callback when a task is canceled

The DefaultTask implementation is modified:
* TaskEvent are published when start() and finish() are called (synchronous task)
* TaskEvent are published just before the runnable is started and after it is  finished
* A try finally statement is used in the async case to ensure a taskevent is sent even in the case of job execution failure
* And event is called when cancel is called, and:
  * Async job : future.cancel(true) is called, + the custom cancel callback is called, if set
  * Sync job : the custom cancel callback is called, if set

* adds a test to check whether many parallel tasks send an event when they are starting and when they are done
* documentation

See also : forum post describing the functionality: https://forum.image.sc/t/demo-and-proposal-new-progress-bars-for-fiji/64956

Draft PR for full functionality will be done in scijava-ui-swing, imagej-legacy and batch-processor

### DIFF
--- a/src/main/java/org/scijava/task/DefaultTask.java
+++ b/src/main/java/org/scijava/task/DefaultTask.java
@@ -36,11 +36,32 @@ import org.scijava.task.event.TaskEvent;
 import org.scijava.thread.ThreadService;
 
 /**
- * Default implementation of {@link Task}. It launches code via the linked
- * {@link ThreadService}, and reports status updates via the linked
- * {@link EventService}.
+ * Default implementation of {@link Task}. Throughout the task (or job),
+ * {@link Task#setProgressValue(long)} can be called to inform
+ * how the job is progressing.
  *
- * @author Curtis Rueden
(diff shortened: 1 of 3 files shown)


NicoKiaru:
I think a `Runnable getCancelCallback()` is missing in the Task interface if one want to concatenate callbacks ( imagine adding a user confirmation dialog in swing before actually calling cancel).

NicoKiaru:
Thanks! I need to work and clean the 3 other PR now ;-)

PR #433, opened 2022-04-10
Prevent TextFormat plugins from claiming support for unknown / unsupported files

ngladitz:
This is meant to fix one of the issues I found in fiji/fiji#303.
Without this change SCIFIO ends up using the `TextIOPlugin` for otherwise unsupported binary files.
I tried extending the existing test to cover the change and document the intend but I am not familiar with the general architecture and don't do a lot of Java so I might be missing if this has unintended consequences or might be unacceptable for other reasons.

### DIFF
--- a/src/main/java/org/scijava/text/AbstractTextFormat.java
+++ b/src/main/java/org/scijava/text/AbstractTextFormat.java
@@ -32,6 +32,7 @@ package org.scijava.text;
 import java.io.File;
 
 import org.scijava.plugin.AbstractHandlerPlugin;
+import org.scijava.util.FileUtils;
 
 /**
  * Abstract superclass of {@link TextFormat} implementations.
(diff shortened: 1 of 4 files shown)


ctrueden:
@ngladitz Thanks. Could you please give me a minimal, reproducible example so that I can test this fix?

ngladitz:
Thank you for picking this up so quickly.

I've extended the test to cover the change but if you mean for manual verification what I did is e.g.:
- extract `test1.obf` from `test.zip` (https://github.com/fiji/fiji/files/8449459/test1.zip)
- open it in Fiji via `File -> Open..`

Without the change I get a dialog with title "Open Text URL" that contains:
Image

With the change I get a console warning "No appropriate format found: ..." and the expected Bio-Formats import dialog.

There seem to be a lot of additional issues in this context that all chain up here but I'll try to tackle them one by one at least as far as they impact fiji/fiji#303.

e.g. the issue plays out differently in Fiji depending on the content of the file.
If Fiji thinks the MarkdownTextFormat processed HTML wrapped content of the file (which may be a large binary file) may be an URL it tries to open that which I assume is probably not supposed to happen either but is presumably an issue located elsewhere.

mkitti:
@ctrueden If the only thing we are waiting for is creating a test case, and large HDF5 files seem to trigger this bug, perhaps we should just create a a large HDF5 file?

ctrueden:
Ach, I'm sorry @ngladitz, I didn't realize (or didn't remember) about this PR when following up on fiji/fiji#303. It looks like my change with aa7fe2e241c9e5dcca2ffad7945475aca47dc72c is the same fix as your PR here. I'm going to close this PR, since I think the problem is resolved now.

PR #435, opened 2022-06-17
Improve array -> `String` conversions from `DefaultConverter`

gselzer:
This PR fixes the issue described in imagej/napari-imagej#65 by improving the conversion of arrays to `Strings`.

We make a few changes:
* Add case logic into `DefaultConverter` for the conversion from array types to `String`s. Previously, for an array the conversion would output `Object.toString()`, which could not be converted back into the array. e.g. for `new double[] {1, 2, 3}`, we'd get something like `[D@1234567`. Now we get `1.0,2.0,3.0`. 
* Add functionality in `ArrayUtils.toCollection` to split a `String` by a delimiter (which is always just "," :sweat_smile:) before shoving it into an `ObjectArray`. This matches previous behavior when there are no commas, but would break previous behavior if there are commas...

Concerns:
* Is the converted `String` specific enough? Do we need to add the component type into the `String` somehow?
* Might this break backwards compatibility, for either change?
* Can/should we test more types?

This *should* be enough to close imagej/napari-imagej#65 once this change makes its way into a release.


gselzer:
On second thought, why do we shove so much case logic into `DefaultConverter`?  Why aren't these all split into separate `Converter`s?

ctrueden:
> why do we shove so much case logic into `DefaultConverter`? Why aren't these all split into separate `Converter`s?

Hysterical raisins. There didn't used to be modular converters, and at one point I split a bunch of logic into separate plugins, but was not able to fully split everything. Please feel free to do it, as your time allows!

gselzer:
Sounds good. It's `Priority.EXTREMELY_LOW` as far as I'm concerned, so maybe I'll just make an issue about it.

src/test/java/org/scijava/module/ModuleServiceTest.java
```diff
@@ -158,6 +159,66 @@ public void testGetSingleInput() throws ModuleException {
...
+			else assertEquals(expected, actual);
+		}
+	}
+
+	private void assertArrayEquality(Object arr1, Object arr2) {
+		// Ensure that both Objects are arrays of the same type!
+		assertEquals(arr1.getClass(), arr2.getClass());
+		assertTrue(arr1.getClass().isArray());
+
+		// We must check primitive arrays as they cannot be cast to Object[]
```
ctrueden, line 203: The `byte[]` case is missing.

src/test/java/org/scijava/module/ModuleServiceTest.java
```diff
@@ -158,6 +159,66 @@ public void testGetSingleInput() throws ModuleException {
...
+			double[].class, new double[] { 1., 2., 3. } //
+		});
+		parameterizations.add(new Object[] { //
+			Double[].class, new Double[] {} //
+		});
+		parameterizations.add(new Object[] { //
+			Double[].class, new Double[] { 1., 2., 3. } //
+		});
+
+		for (Object[] params : parameterizations) {
```
ctrueden, line 179: It looks like `params[0] == params[1].getClass()` for every element. So then, can't this be simplified to:
```java
List<Object> parameterizations = new ArrayList<>();
parameterizations.add(new double[] {});
parameterizations.add(new double[] {1., 2., 3.});
parameterizations.add(new Double[] {});
parameterizations.add(new Double[] {1., 2., 3.});
```
and then:
```java
Object expected = params;
Class<?> type = expected.getClass();
```
?

And maybe change the name from `parameterizations` to just `objects`?

src/test/java/org/scijava/module/ModuleServiceTest.java
```diff
@@ -158,6 +159,66 @@ public void testGetSingleInput() throws ModuleException {
...
+			Class<?> type = (Class<?>) params[0];
+			Object expected = params[1];
+			// Get a ModuleItem of the right type
+			MutableModule m = new DefaultMutableModule();
+			m.getInfo().addInput(new DefaultMutableModuleItem<>(m, "a", type));
+			@SuppressWarnings("unchecked")
+			final ModuleItem<Object> item = (ModuleItem<Object>) moduleService
```
ctrueden, line 186: Casting to `ModuleItem<Object>` here is wrong. It must be `ModuleItem<?>` because we don't actually know the module item's type. The type is not actually `Object` in any of the four concrete cases declared above.

I understand that this was done so that the following `save` call would compile, but the correct way to do this would be to have a private helper function parameterized on `T` that accepts the `expected` as a `T` argument, so that you have a type variable to declare the return value of `getSingleInput` as a `ModuleItem<T>`, and then subsequently use it with `save`.

src/test/java/org/scijava/module/ModuleServiceTest.java
```diff
@@ -158,6 +159,66 @@ public void testGetSingleInput() throws ModuleException {
...
+		}
+	}
+
+	private void assertArrayEquality(Object arr1, Object arr2) {
+		// Ensure that both Objects are arrays of the same type!
+		assertEquals(arr1.getClass(), arr2.getClass());
+		assertTrue(arr1.getClass().isArray());
+
+		// We must check primitive arrays as they cannot be cast to Object[]
+		if (arr1 instanceof boolean[]) //
```
ctrueden, line 204: Using `{ }` around the conditionals here would better match the style elsewhere in SciJava Common, e.g. in `ArrayUtils`.

src/main/java/org/scijava/convert/DefaultConverter.java
```diff
@@ -168,10 +169,20 @@ public <T> T convert(final Object src, final Class<T> dest) {
 		}
 		if (saneDest == String.class) {
 			// destination type is String; use Object.toString() method
-			final String sValue = src.toString();
-			@SuppressWarnings("unchecked")
-			final T result = (T) sValue;
-			return result;
+			if (src.getClass().isArray()) {
+				final String elementString = ArrayUtils.toCollection(src).stream() //
```
ctrueden, line 173: Nice use of `ArrayUtils.toCollection`. :+1:

src/main/java/org/scijava/convert/DefaultConverter.java
```diff
@@ -168,10 +169,20 @@ public <T> T convert(final Object src, final Class<T> dest) {
...
+					.map(object -> convert(object, String.class)) //
+					.collect(Collectors.joining(","));
```
ctrueden, line 175: Might be more human-readable to use `", "` (with trailing space) rather than `","`.

src/main/java/org/scijava/convert/DefaultConverter.java
```diff
@@ -168,10 +169,20 @@ public <T> T convert(final Object src, final Class<T> dest) {
...
+				@SuppressWarnings("unchecked")
+				final T result = (T) elementString;
```
ctrueden, line 177: I think the array should be prefixed by `{` and suffixed by `}`. Otherwise:
```java
String[][] s = { {"hello", "there"}, {"goodbye", "now"} };
```
becomes:
```
hello, there, goodbye, now
```
And it's not obvious that it was once nested arrays.

And even then, other cases will be hosed upon reconstitution, such as strings containing `,` or `{` or `}` characters. We would need to perform escaping/unescaping to avoid that.

Based on the unit test below, I assume the primary use case here is array persistence, so then the escaping is necessary, so that strings converted from arrays are convertible back to an equivalent array object afterward. Right? In that case, it would be better to support direct persistence of properly typed arrays, rather than feeding through the converter framework here, don't you think?

ctrueden (changes requested):
I'm not sure we want to do this. If we do, then I think we need a more robust stringification of the array data. In that case, I would recommend escaping each string into a YAML-compliant form, since YAML solves essentially the same problem of how to express a properly typed array in string form.

ctrueden:
@gselzer To answer your specific questions:

> Is the converted String specific enough? Do we need to add the component type into the String somehow?

I think it is OK not to. The conversion is allowed to be lossy. If we baked in the component type of the array, it would make reconstruction from string more rigorous, but also more complicated.

> Might this break backwards compatibility, for either change?

I don't think it breaks backwards compatibility in a broad sense, but specifically for strings passed to `ArrayUtils.toCollection`, it does. Previously, strings would be wrapped into singleton collections. We should not change this behavior.

> Can/should we test more types?

I pointed out some cases where the conversion is currently problematic. We should test those cases. In particular, for nested arrays, and strings containing special characters that need escaping, we should test.

src/main/java/org/scijava/util/ArrayUtils.java
```diff
@@ -120,6 +122,12 @@ public static Collection<?> toCollection(final Object value) {
 		if (value instanceof Object[]) {
 			return new ObjectArray<>((Object[]) value);
 		}
+		if (value instanceof String) {
+			String[] arr = ((String) value).split("\\s*,\\s*");
```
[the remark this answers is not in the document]
gselzer: > This change does not abide by the intent of the `toCollection` method.

Yes, you are right. This, and my question above about why we are thowing so much into existing plugins/classes, made me decide to do a huge refactor.  That refactor removes any alterations to `ArrayUtils`.

src/test/java/org/scijava/module/ModuleServiceTest.java
```diff
@@ -42,6 +40,9 @@
 import org.junit.Before;
 import org.junit.Test;
 import org.scijava.Context;
+import org.scijava.util.ObjectArray;
+
+import static org.junit.Assert.*;
```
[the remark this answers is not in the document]
gselzer: Dang, I thought I had fixed it before. IntelliJ has a separate setting for converting `static` imports to wildcard imports, which I had not changed.

Fixed now.

src/main/java/org/scijava/convert/DefaultConverter.java
gselzer: Thanks, fixed.

src/main/java/org/scijava/convert/DefaultConverter.java
gselzer: Great, thanks. I wrote a ton of unit tests that also test nested arrays.

> Based on the unit test below, I assume the primary use case here is array persistence, so then the escaping is necessary, so that strings converted from arrays are convertible back to an equivalent array object afterward. Right? In that case, it would be better to support direct persistence of properly typed arrays, rather than feeding through the converter framework here, don't you think?

Maybe, but would that not be more work?

src/test/java/org/scijava/module/ModuleServiceTest.java
gselzer: Yes, all good revisions. Just fixed them.

src/test/java/org/scijava/module/ModuleServiceTest.java
gselzer: Fixed!@

src/test/java/org/scijava/module/ModuleServiceTest.java
gselzer: Thanks, I fixed it in the most recent commit.

src/test/java/org/scijava/module/ModuleServiceTest.java
gselzer: Not anymore!

gselzer:
> I'm not sure we want to do this.

Do you mean that we might want to close the PR?

> If we do, then I think we need a more robust stringification of the array data. In that case, I would recommend escaping each string into a YAML-compliant form, since YAML solves essentially the same problem of how to express a properly typed array in string form.

I'm not sure that I understand what you mean by a "YAML-compliant form".  Can you elaborate on what you mean? I do agree, looking back, that we should probably be using some sort of brackets if possible; I'm not sure why I did not do that before.

> > Is the converted String specific enough? Do we need to add the component type into the String somehow?
> 
> I think it is OK not to. The conversion is allowed to be lossy. If we baked in the component type of the array, it would make reconstruction from string more rigorous, but also more complicated. It would also make converted arrays harder for humans to read.

:+1:

> > Might this break backwards compatibility, for either change?
> 
> I don't think it breaks backwards compatibility in a broad sense, but specifically for strings passed to `ArrayUtils.toCollection`, it does. Previously, strings would be wrapped into singleton collections. We should not change this behavior.

Yeah, you are right. I removed the changes made to that method.

> > Can/should we test more types?
> 
> I pointed out some cases where the conversion is currently problematic. We should test those cases. In particular, for nested arrays, and strings containing special characters that need escaping, we should test.

Thanks, I tested nested arrays, empty arrays, is there anything else that we should be testing?

gselzer:
Thanks again for taking a look @ctrueden

ctrueden (changes requested):
Thanks @gselzer! I think this is getting much closer! You asked about what other things to test... right now I think arrays of strings with `,` or `{` or `}` symbols is at risk of failing to convert successfully back from their stringified form. I outline a potential way of escaping the strings with backslashes. Another way would be the HTML-style way with ampersand codes, which would make it easier to unescape the stringified data. Either way, I suggest you take a look at Parsington, because it can parse nested expressions with curly braces by default. The main compromise you'd need to adopt if you use Parsington is that strings would need to be surrounded by double quotes, Java style.

src/main/java/org/scijava/convert/StringToArrayConverter.java
```diff
@@ -0,0 +1,178 @@
...
+		return saneSrc == String.class && saneDest.isArray();
+	}
+
+	@Override
+	public boolean canConvert(final Object src, final Class<?> dest) {
+		if (!canConvert(src.getClass(), dest)) return false;
+		String srcString = (String) src;
+		if (!(srcString.startsWith("{") && srcString.endsWith("}"))) return false;
+		List<String> components = elements((String) src);
+		return components.size() == 0 || convertService.supports(components.get(0),
```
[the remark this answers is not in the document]
gselzer: Good point, thanks.

> But I think it's an acceptable compromise, since the only way to be completely sure would be to actually do the full conversion here.

We *could* consider doing the conversion here, though, and caching it for later in a `WeakHashMap`...but I feel like that could have its own issues, e.g. with repeated calls to `convert` returning the same object...so I'm not going to implement this.

src/main/java/org/scijava/convert/StringToArrayConverter.java
```diff
@@ -0,0 +1,178 @@
...
+			dest.getComponentType());
+	}
+
+	@Override
+	public Object convert(Object src, Type dest) {
+		final Type componentType = Types.component(dest);
+		if (componentType == null) throw new IllegalArgumentException(dest +
```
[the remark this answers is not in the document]
gselzer: Thanks, fixed in 8203b522.

src/main/java/org/scijava/convert/ArrayToStringConverter.java
```diff
@@ -0,0 +1,90 @@
...
+
+	@Override public boolean canConvert(final Object src, final Class<?> dest) {
+		if (!canConvert(src.getClass(), dest)) return false;
+		if (Array.getLength(src) == 0) return true;
+		return convertService.supports(Array.get(src, 0), dest);
+	}
+
+	@Override public Object convert(Object src, Type dest) {
+		final String elementString = ArrayUtils.toCollection(src).stream() //
+				.map(object -> convertService.convert(object, String.class)) //
```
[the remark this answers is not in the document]
gselzer: @ctrueden does Parsington have functionality to turn data into a parseable string? i.e. could it do the Array -> String conversion?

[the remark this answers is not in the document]
ctrueden: @gselzer No, it doesn't do stringification of Java data structures. It is for the other direction: string to syntax tree. So unfortunately we'll still need to write some logic here to e.g. escape strings appropriately for harmony with Java/Parsington syntax.

src/test/java/org/scijava/module/ModuleServiceTest.java
```diff
@@ -43,6 +41,12 @@
 import org.junit.Test;
 import org.scijava.Context;
 
+import static org.junit.Assert.assertArrayEquals;
```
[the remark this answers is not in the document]
gselzer: Fixed in 24434872

src/test/java/org/scijava/module/ModuleServiceTest.java
```diff
@@ -158,6 +162,70 @@ public void testGetSingleInput() throws ModuleException {
...
+		objects.add(new Object[] { new double[] { 1., 2., 3. } });
+		objects.add(new Object[] { new Double[] {} });
+		objects.add(new Object[] { new Double[] { 1., 2., 3. } });
+
+		for (Object[] params : objects) {
+			saveParam(params[0]);
+		}
+	}
+
+	private <T> void saveParam(T object) {
```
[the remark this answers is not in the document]
gselzer: Fixed in 24434872

src/test/java/org/scijava/module/ModuleServiceTest.java
```diff
@@ -158,6 +162,70 @@ public void testGetSingleInput() throws ModuleException {
...
+	public void testSaveAndLoad() {
+
+		List<Object[]> objects = new ArrayList<>();
+		objects.add(new Object[] { new double[] {} });
+		objects.add(new Object[] { new double[] { 1., 2., 3. } });
+		objects.add(new Object[] { new Double[] {} });
+		objects.add(new Object[] { new Double[] { 1., 2., 3. } });
+
+		for (Object[] params : objects) {
+			saveParam(params[0]);
```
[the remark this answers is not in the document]
gselzer: Fixed in 24434872

src/main/java/org/scijava/convert/ArrayToStringConverter.java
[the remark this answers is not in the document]
gselzer: Okay, that's what I figured. Just making sure.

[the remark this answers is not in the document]
gselzer: I think this is all fixed now.

gselzer:
@ctrueden I think we are in a much better place now with Parsington, thanks for the suggestion!

ctrueden:
@gselzer I added a couple more asserts, checking that the result strings converted from arrays are what we expect, before the tests continue and convert them back into arrays again. I wanted to be totally sure that strings with more challenging characters are escaped correctly.

ctrueden:
@gselzer Turns out there were still problems after I merged this PR. Now fixed in 42db182fa90ebe4fc79163713b8cf5ac7ee9b143 and 497eff6eca3ee9687acddd9453859fef40071390.

gselzer:
Thanks for those fixed @ctrueden. Admittedly I went through and made the changes pretty quick.

PR #437, opened 2022-07-01
Add a StringToNumberConverter

gselzer:
This PR adds a `StringToNumberConverter`, a dedicated `Converter` that converts from `String`s to:
* numerical primitives
* `Number`
* `Number`'s subclasses

This is useful for imagej/imagej-ops#644

### DIFF
--- a/src/main/java/org/scijava/convert/StringToNumberConverter.java
+++ b/src/main/java/org/scijava/convert/StringToNumberConverter.java
@@ -1,0 +1,56 @@
+
+package org.scijava.convert;
+
+import org.scijava.plugin.Plugin;
+import org.scijava.util.Types;
+
+/**
+ * Converts a {@link String} to a {@link Number}. Currently handles all boxed
+ * and unboxed primitives, along with the Numbers. In particular,
+ * {@link String}s converted {@link Number}s are just {@link Double}s, as this
+ * features the broadest range of integers.
+ *
+ * @author Gabriel Selzer
+ */
+@Plugin(type = Converter.class)
+public class StringToNumberConverter extends AbstractConverter<String, Number> {
+
+	@Override
+	@SuppressWarnings("unchecked")
+	public <T> T convert(Object src, Class<T> dest) {
+		// ensure type is well-behaved, rather than a primitive type
+		Class<T> saneDest = Types.box(dest);
+		if (!(src instanceof String)) throw new IllegalArgumentException(
+			"Expected src to be a String but got a " + src.getClass());
+		if (!(Number.class.isAssignableFrom(saneDest)))
+			throw new IllegalArgumentException(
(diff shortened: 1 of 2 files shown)


src/test/java/org/scijava/convert/StringToNumberConverterTest.java
```diff
@@ -0,0 +1,112 @@
...
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * Tests {@link StringToNumberConverter}
+ *
+ * @author Gabriel Selzer
+ */
+public class StringToNumberConverterTest {
```
ctrueden, line 13: Needs tests that non-numerically-formatted strings successfully report `false` for `canConvert`.

src/main/java/org/scijava/convert/StringToNumberConverter.java
```diff
@@ -0,0 +1,56 @@
...
+
+	@Override
+	public Class<String> getInputType() {
+		return String.class;
+	}
+
+	@Override
+	public boolean canConvert(Object src, Class<?> dest) {
+		if (!Types.isAssignable(src.getClass(), String.class)) return false;
+		return Types.isAssignable(Types.box(dest), Number.class);
```
ctrueden, line 54: Needs to check that the string is actually a valid numeric format. The best way to do that is, sadly, to actually do the conversion. But that's OK.

ctrueden (changes requested):
As written, this will cause the convert service to erroneously report strings like `"Hello"` as convertible to `Number` subclasses. We need to be selective in which strings we allow here, no? In the same way we are being selective e.g. with unwrapping `ROITree` objects?

gselzer:
Thanks for the feedback @ctrueden!

Both your concerns should be fixed in 5c160d2

src/main/java/org/scijava/convert/StringToNumberConverter.java
```diff
@@ -51,6 +50,22 @@ public Class<String> getInputType() {
 	@Override
 	public boolean canConvert(Object src, Class<?> dest) {
 		if (!Types.isAssignable(src.getClass(), String.class)) return false;
-		return Types.isAssignable(Types.box(dest), Number.class);
+		// The only way to know if the conversion is valid is to actually do it.
+		try {
+			String srcString = (String) src;
+			sane(dest).getConstructor(String.class).newInstance(srcString);
```
ctrueden, line 56: This reflection, while slick, concerns me w.r.t. the Java 11+ restrictions on reflection. But it's fine for now, we can always fix later as needed.

gselzer: Yes, you are right. `DefaultConverter` does the same thing, though, so I think you are right to fix these later.

PR #438, opened 2022-07-07
Set jna.library.path to java.library.path if not otherwise set

mkitti:
If `jna.library.path` is not set, set it to be the  same as `java.library.path`.

This should fix the issue with version libraries on Linux and macOS, allowing libblosc.so.1 to be dynamically linked by JBlosc.

### DIFF
--- a/src/main/java/org/scijava/platform/DefaultPlatformService.java
+++ b/src/main/java/org/scijava/platform/DefaultPlatformService.java
@@ -59,8 +59,8 @@ public final class DefaultPlatformService extends
 		//For ImageJ or Fiji, java.library.path = ".../Fiji.app/lib/win64:.../Fiji.app/mm/win64"
 		String java_library_path = System.getProperty("java.library.path");
 		
-		//Used by Java Native Access for JBlosc, sync with java.library.path
-		if(System.getProperty("jna.library.path") == null) {
+		//Allow Java Native Access to search java.library.path for versioned shared libraries on Linux
+		if(java_library_path != null && System.getProperty("jna.library.path") == null) {
 			System.setProperty("jna.library.path", java_library_path);
 		}
 	}
--- a/src/main/java/org/scijava/platform/DefaultPlatformService.java
+++ b/src/main/java/org/scijava/platform/DefaultPlatformService.java
@@ -54,6 +54,16 @@ import org.scijava.service.Service;
 public final class DefaultPlatformService extends
 	AbstractSingletonService<Platform> implements PlatformService
 {
+	
+	static {
+		//For ImageJ or Fiji, java.library.path = ".../Fiji.app/lib/win64:.../Fiji.app/mm/win64"
+		String java_library_path = System.getProperty("java.library.path");
+		
+		//Used by Java Native Access for JBlosc, sync with java.library.path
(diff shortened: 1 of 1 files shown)


ctrueden:
@mkitti Sorry for the slow reply on this. I want to merge this, but I'd prefer some more general comments on what this accomplishes, with the oblique mentions of Fiji.app. This library is more general than ImageJ2 and Fiji, after all. The comment should say why setting jna.library.path to match java.library.path is a useful thing to do.

mkitti:
The heart of the issue is that currently the FIJI installer which only sets java.library.path and does not set jna.library.path.

JNA will only look in jna.library.path for versioned shared libraries such as libblosc.so.1. Within java.library.path it will only look for libblosc.so.

On macOS or Linux, JBlosc will happily find libblosc.so.1 within system folders since they are on the jna.library.path by default. Within FIJI, JBlosc will not look in Fiji.app/lib for JBlosc because it is not on the jna.library.path.

Thus pull request will enable JNA apps within FIJI to act like JNA apps outside of FIJI. They will both be able to locate libblosc.so.1 whether it is installed within FIJI or by the system.

Otherwise to include binaries in FIJI I have to compile a special version for FIJI that links to libblosc.so. That binary will not work on Ubuntu Linux which only bundles libblosc.so.1 I n the primary libblosc package. libblosc.so is only included libblosc-dev.

A specific application of the issue is coordination between JBlosc and the HDF5 Blosc plugin.

Binaries within FIJI should be able to work outside of FIJI. For this to happen they should link to versioned libraries like libblosc.so.1 rather than unversioned dev libraries like libblosc.so.

This PR allows JNA and native libraries to link against the same versioned library (e.g. libblosc.so.1)

mkitti:
Specifically:


JNA initializes the search path to `jna.library.path` in the private method `NativeLibrary.loadLibrary` (not the deprecated public API):
https://cs.github.com/java-native-access/jna/blob/0444bf937cc226f079a8b2698ad193abf86d50d8/src/com/sun/jna/NativeLibrary.java?q=initPaths#L187

The search for variants begins later in that method if an `UnsatisfiedLinkError` is thrown:
https://cs.github.com/java-native-access/jna/blob/0444bf937cc226f079a8b2698ad193abf86d50d8/src/com/sun/jna/NativeLibrary.java?q=initPaths#L232-L262

The `matchLibrary` function implements variant searching on Linux:
https://cs.github.com/java-native-access/jna/blob/0444bf937cc226f079a8b2698ad193abf86d50d8/src/com/sun/jna/NativeLibrary.java?q=initPaths#L824

mkitti:
Here's an image I previously posted on [imagesc Zulip](https://imagesc.zulipchat.com/#narrow/stream/327237-SciJava/topic/JHDF5.20and.20native-lib-loader/near/285502883).

image

My current problem is that I need to link H5ZBlosc, the HDF5 plugin for Blosc, against either `libblosc.so.1` or `libblosc.so`. If I link against `libblosc.so.1`, it will work on most Linux operating systems but not FIJI. If I link against `libblosc.so`, it will work on FIJI but not most Linux operating systems unless development versions are installed.

To resolve this, I would like to change libblosc.so to libblosc.so.1 in FIJI. This requires that `jna.library.path` be set as described above.

PR #447, opened 2022-12-12
Fix priority of `LoggerPreprocessor`

### LINKED ISSUE #446: `Logger` parameters can't be used while initializing `DynamicCommand`s
While using a `LogService` parameter for logging during the `initialize` method of an initializable `DynamicCommand` works just fine, the same cannot be done with a `Logger` parameter, as it throws a NullPointerException.

I assume the reason for this is the different priorities of `LoggerPreprocessor` vs. `ServicePreprocessor`:

https://github.com/scijava/scijava-common/blob/f1f666a8aded15e05660aece5707cd0f8b12521d/src/main/java/org/scijava/module/process/LoggerPreprocessor.java#L47-L48

https://github.com/scijava/scijava-common/blob/f1f666a8aded15e05660aece5707cd0f8b12521d/src/main/java/org/scijava/module/process/ServicePreprocessor.java#L62-L63

... and can likely be solved by increasing the priority of the former. I am unsure though if this might have any unwanted side effects elsewhere.

@ctrueden, @maarzt: if you agree, I'll test and file a pull request.

imagejan:
`Logger` parameters should be processed before `InitPreprocessor`, so that plugins can use a `Logger` in their initializer.

Closes #446.

### DIFF
--- a/src/main/java/org/scijava/module/process/LoggerPreprocessor.java
+++ b/src/main/java/org/scijava/module/process/LoggerPreprocessor.java
@@ -29,6 +29,7 @@
 
 package org.scijava.module.process;
 
+import org.scijava.Priority;
 import org.scijava.log.LogService;
 import org.scijava.log.Logger;
 import org.scijava.module.Module;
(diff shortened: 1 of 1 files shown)


ctrueden:
@imagejan Thanks! So the change to `VERY_HIGH` ends up moving the `LoggerPreprocessor` to *before* the `InitPreprocessor` (which is `HIGH`) but still *after* the `ServicePreprocessor` (which is `2*VERY_HIGH`). That's intended, right?

imagejan:
Thanks for merging, @ctrueden!
Exactly, moving `LoggerPreprocessor` priority in between `ServicePreprocessor` and `InitPreprocessor` was intended.

PR #449, opened 2022-12-13
Only convert to array types when it is possible

### LINKED ISSUE #448: ConvertService eagerly converts stuff to arrays
`ij.convert().supports(java.lang.System, boolean[])` returns `true` :unamused: 

This is because [`DefaultConverter.canConvert`](https://github.com/scijava/scijava-common/blob/9564503c70cf49a21fa15b988552a3204f801181/src/main/java/org/scijava/convert/DefaultConverter.java#L295) makes no checks on the source type if the destination type is an array type.

gselzer:
This closes #448.

Instead of reporting that `DefaultConverter` can convert *anything* to an array, we only report that we can convert things that `ArrayUtils.toCollection` can actually convert into a `Collection`.

One other concern is whether the case logic might lead to action-at-a-distance errors.

### DIFF
--- a/src/main/java/org/scijava/convert/DefaultConverter.java
+++ b/src/main/java/org/scijava/convert/DefaultConverter.java
@@ -292,7 +292,17 @@ public class DefaultConverter extends AbstractConverter<Object, Object> {
 	public boolean canConvert(final Class<?> src, final Type dest) {
 
 		// Handle array types, including generic array types.
-		if (isArray(dest)) return true;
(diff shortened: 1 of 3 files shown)


gselzer:
@ctrueden how do these changes fare with backwards compatibility?

ctrueden:
> how do these changes fare with backwards compatibility?

I don't foresee any problems! The tests all passed, so it must be bug-free! :trophy:

PR #455, opened 2023-03-14
DefaultIOService: Add concrete methods to resolve Location

ecmagnuson:
Moved implementation of getOpener(String) and getSaver(String) to DefaultIOService so they can use LocationService, instead of forcing FileLocation.  

closes https://github.com/imagej/tutorials/issues/101.


hinerm (approved):
Looks good to me! Wow, even had the `LocationService` already available in the `DefaultIOService`. Nice!

PR #462, opened 2023-05-19
Find scripts in base script directory

gselzer:
This PR allows scripts to be found in the base script directory.

### DIFF
--- a/src/main/java/org/scijava/script/ScriptFinder.java
+++ b/src/main/java/org/scijava/script/ScriptFinder.java
@@ -41,6 +41,7 @@ import java.util.Set;
 
 import org.scijava.AbstractContextual;
 import org.scijava.Context;
+import org.scijava.MenuEntry;
 import org.scijava.MenuPath;
 import org.scijava.log.LogService;
 import org.scijava.plugin.Parameter;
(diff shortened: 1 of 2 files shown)


hinerm:
This works, functionally, in principle. In practice, when I drop this in Fiji the discovered scripts aren't actually accessible via the menus because there is no "Scripts" root menu path:
image

What about just putting them in `/plugins/Scripts` anyway? since that's where they go on the file system?

hinerm:
> In practice, when I drop this in Fiji the discovered scripts aren't actually accessible via the menus because there is no "Scripts" root menu path:

OK this was me being dumb and forgetting that everything, including scripts, need to have an underscore (`_`) in the file name to be placed into the GUI menus.

So yeah, this fix works. But I don't love that it creates a new menu entry called "Scripts" because.. there are tons of scripts that are NOT going into that directory (most of them)

In a local Fiji installation the `scripts` directory is in the `plugins` directory, so collecting them into the `Plugins` menu or submenu thereof seems reasonable?

src/main/java/org/scijava/script/ScriptFinder.java
```diff
@@ -178,8 +179,10 @@ private int createInfos(final List<ScriptInfo> scripts, final Set<URL> urls,
 			// friendlyPath = "File/Import/Movie File..."
 			// menuPath = File > Import > Movie File...
 
-			// NB: Ignore base-level scripts (not nested in any menu).
-			if (menuPath.size() == 1) continue;
+			// Place base-level scripts in the "Scripts" submenu
+			if (menuPath.size() == 1){
+				menuPath.add(0, new MenuEntry("Scripts"));
```
hinerm, line 184: We don't want a new top-level "Scripts" directory. I suggest "Plugins > Scripts" since that matches their location on disk.

hinerm (changes requested):

hinerm (approved):

PR #465, opened 2023-07-12
Manage the active UI in a better way

### LINKED ISSUE #460: UIService.showUI() shows multiple instances of the same UI
I found this with the ImageJ2 UI in napari-imagej:

image

Ideally, `UIService.show()` would check if there is already a UI visible, and then, if it is, ensure that the application frame is also visible. This is what I was doing [in napari-imagej](https://github.com/imagej/napari-imagej/blob/2f4c808ebf3dca079a23821bc6b0c0787cc55bab/src/napari_imagej/widgets/menu.py#L57), however it *is* a little hacky (why would the ImageJ2 UI be visible and the application frame not be?)

Note that in napari-imagej, I also kill all `WindowListener`s, and add my own that [hides the window](https://github.com/imagej/napari-imagej/blob/2f4c808ebf3dca079a23821bc6b0c0787cc55bab/src/napari_imagej/widgets/menu.py#L113) when the user presses the close button, instead of actually closing the window.

ctrueden:
Thanks to the HeadlessUI, we now always have at least one UI available to SciJava Common. This commit introduces the concept internally of an *active* UI: the one that will be used when UI-centric operations are performed via the UIService (rather than directly on a UserInterface).

The logic is as follows:

* If an active UI already exists and is currently visible, use it again.
* Otherwise, make the first visible UI the active one, and use it.
* If none, make the default UI the active one, and use it.

This PR also fixes a regression, introduced in #461, where calling `uiService.show(thing)` would also cause the default UI to become visible. Previously, it would only show that thing using the default UI, but the main window would not appear, and the UI as a whole would still be considered not visible. (To be fair: the "regression" here was explicitly stated as an intended change in that PR—but after discussion and reflection, we decided we don't want to change this behavior after all.)

~~Might also fix #460?~~ Edit: Nope, doesn't.

### DIFF
--- a/src/main/java/org/scijava/ui/DefaultUIService.java
+++ b/src/main/java/org/scijava/ui/DefaultUIService.java
@@ -117,6 +117,9 @@ public final class DefaultUIService extends AbstractService implements
 	/** The default user interface to use, if one is not explicitly specified. */
 	private UserInterface defaultUI;
 
+	/** The last UI used when performing UI operations via the service. */
+	private UserInterface activeUI;
+
 	/**
 	 * When true, {@link #isHeadless()} will return true regardless of the value
 	 * of the {@code java.awt.headless} system property. When false, {@link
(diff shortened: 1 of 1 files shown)


gselzer (approved):
I have no concerns, this seems like a good change to me!

elevans:
This works great. I can display `net.imagej.Dataset` images without displaying the ImageJ UI in PyImageJ.

PR #477, opened 2024-05-09
Add text-based property util

hinerm:


hinerm:
@ctrueden is this what you had in mind?

PR #479, opened 2024-06-27
Ignore empty Strings during String to FileArray conversion

stefanhahmann:
* The current state of the code is that an empty String, is converted to a new File("") object initialized with an empty String
* A File object instantiated by this will be a File object with an absolute Path that is equivalent to the Path from which the application was started
* The result of this is a bit unpredictable
* This PR thus suggests that an empty String should not be converted to File object at all and not be added to the output


imagejan (approved):

ctrueden:
@stefanhahmann Thanks!

PR #480, opened 2024-06-27
Add FileWidget.FILE_AND_DIRECTORY_STYLE

tischi:


NicoKiaru:
Quick note to mention that the documentation in imagej.net needs to be updated once the PR is merged:

https://imagej.net/scripting/parameters

NicoKiaru:
Also, instead of "both", maybe something like "file_or_directory", since "both" isn't very clear out of context.
Last, how does this behave with a file array parameter ?

tischi:
I am using "both" because that's what is [already used for the FileListWidget](https://github.com/scijava/scijava-common/blob/d4ab47a1d41d9f7c0e593afb6a7959119c14249e/src/main/java/org/scijava/widget/FileListWidget.java#L53).

NicoKiaru:
Ah great! Then the FileArray widget is already working. Good to know!

src/main/java/org/scijava/ui/UserInterface.java
```diff
@@ -163,6 +163,7 @@ default File chooseFile(final File file, final String style) {
 		// TODO use a utility class for style handling, e.g. StyleUtils.isStyle(style, ...)
 		if (style == null) title = "Choose a file";
 		else if (style.toLowerCase().contains(FileWidget.DIRECTORY_STYLE)) title = "Choose a directory";
+		else if (style.toLowerCase().contains(FileWidget.FILE_AND_DIRECTORY_STYLE )) title = "Choose a file or directory";
 		else if (style.toLowerCase().contains(FileWidget.OPEN_STYLE)) title = "Open";
 		else if (style.toLowerCase().contains(FileWidget.SAVE_STYLE)) title = "Save";
 		else title = "Choose a file";
```
imagejan, line 169: While at it, we might address the TODO comment here:

```suggestion
		if (style == null) title = "Choose a file";
		else if (WidgetStyle.isStyle(style, FileWidget.DIRECTORY_STYLE)) title = "Choose a directory";
		else if (WidgetStyle.isStyle(style, FileWidget.FILE_AND_DIRECTORY_STYLE )) title = "Choose a file or directory";
		else if (WidgetStyle.isStyle(style, FileWidget.OPEN_STYLE)) title = "Open";
		else if (WidgetStyle.isStyle(style, FileWidget.SAVE_STYLE)) title = "Save";
		else title = "Choose a file";
```

(needs the corresponding import for `org.scijava.widget.WidgetStyle`)

imagejan (commented):
Thanks @tischi!

Looks good to me, I just added suggestion to remove one of the TODO comments.

tischi:
Thanks @imagejan, these TODO were not added by me, but what you did LGTM. 

Do I need to accept/commit your changes or can you do it?

ctrueden:
Thanks, all!

PR #481, opened 2024-07-22
Show UI on EDT if required

hinerm:
Closes https://github.com/imagej/imagej-legacy/issues/306


ctrueden:
@hinerm Check out 8e2b30203a0a826955ca780b8b3891b3f972219d and let me know what you think.

PR #482, opened 2024-09-06
Fix duplicate entries in input selectors

hinerm:
This PR represents an alternative to [this PR](https://github.com/imagej/imagej-legacy/pull/171) attempting to fix the duplicates earlier in the framework.

However, this **does** create a problem if the converters themselves consistently create new items. For example, this is what happens when running a trivial script multiple times:

```groovy
//@ImagePlus in1
//@ImagePlus in2

println(in1)
println(in2)
```

image

This could be a bug in particular converters ([ImageTitleToImagePlusConverter](https://github.com/imagej/imagej-legacy/blob/7eac77ab28901975fc6bf0566a899cf1bef06d97/src/main/java/net/imagej/legacy/convert/ImageTitleToImagePlusConverter.java) is a likely suspect) but it does raise a concern of potential behavior for other converters, too.

### DIFF
--- a/src/main/java/org/scijava/widget/AbstractInputHarvester.java
+++ b/src/main/java/org/scijava/widget/AbstractInputHarvester.java
@@ -30,9 +30,10 @@
 package org.scijava.widget;
 
 import java.util.ArrayList;
-import java.util.HashSet;
+import java.util.Collection;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.scijava.AbstractContextual;
 import org.scijava.convert.ConvertService;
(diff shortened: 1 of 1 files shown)


hinerm:
>([ImageTitleToImagePlusConverter](https://github.com/imagej/imagej-legacy/blob/7eac77ab28901975fc6bf0566a899cf1bef06d97/src/main/java/net/imagej/legacy/convert/ImageTitleToImagePlusConverter.java) is a likely suspect)

OK obviously it wasn't the title converter but [DatasetToImagePlusConverter](https://github.com/imagej/imagej-legacy/blob/7eac77ab28901975fc6bf0566a899cf1bef06d97/src/main/java/net/imagej/legacy/convert/DatasetToImagePlusConverter.java#L69) which registers `Datasets` blindly without checking for existing mappings.

A simple fix. We also decided to just not aggressively convert items though, and instead de-duplicate on `Strings`

hinerm:
In conjunction with [these imagej-legacy changes](https://github.com/imagej/imagej-legacy/tree/img-plus-support) this branch now appears to resolve the [Fiji input harvester bug](https://github.com/fiji/fiji/issues/349).

**However**, the `String` de-duplication is of course fragile, and does not address one of the goals in that inputs *will* change after the first run. The only way I'm aware of to satisfy the goal of consistent outputs would be to [aggressively convert input candidates](https://github.com/scijava/scijava-common/tree/convert-widget-inputs). That option is also working with the aforementioned imagej-legacy branch.

So we have to pick one:
a) String comparison 
b) Aggressive conversion

 Functionally I like `b` more but I am worried it will open a whole can of worms in the future as we discover more badly-behaving converters, whereas `a` may prove insufficient in some cases but works in the current obvious, annoying case, and could be expanded by tapping into [Named](https://github.com/scijava/scijava-common/blob/master/src/main/java/org/scijava/Named.java), etc.

hinerm:
OK, decided that this is the least intrusive change (compared to converting) and we can fix the `ImagePlus` title elsewhere if needed

